PR #3, opened 2020-09-04
Don't use goroutines as the ordering is non-deterministic

ibuildthecloud:


rmweir:
https://github.com/rancher/rancher/issues/28179

PR #6, opened 2021-05-24
Add a max timeout bucket

dramich:
Problem:
The bucket being used `rate.Limiter` does not have a maximum timeout so on startup if we process thousands of objects for the same item type this bucket can end up at hours long wait times. It then takes that same amount of time for the bucket to go back to zero. 

Solution:
Wrap the `rate.Limiter` with a max timeout of 2 min

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,8 @@ module github.com/rancher/lasso
 go 1.14
 
 require (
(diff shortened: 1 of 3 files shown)


kinarashah:
https://github.com/rancher/rancher/issues/31999

PR #15, opened 2022-01-12
expose prometheus metrics for controller reconciliation and work queue

aiwantaozi:
Expose work queue metrics and controller metrics, so that the project uses the current library as a dependency could easily obtain monitoring data.
Related Issue: https://github.com/harvester/harvester/issues/1781


rmweir (changes requested):
Should probably use a mechanism like [this](https://github.com/rancher/apiserver/pull/7/files#diff-f9f5c58430e10788eaead208bd14ea2790036edd138c98936fa564a0a9537c61R1-R16) to gate metrics collection behind `CATTLE_PROMETHEUS_METRICS` env var. Also, `go mod tidy` should be ran.

guangbochen:
@rmweir can u please help to merge those two into one PR at https://github.com/rancher/lasso/pull/17 and we are fine to close this one, thanks.

PR #16, opened 2022-02-09
update k8s to 1.23

kinarashah:


controller-runtime/go.mod
```diff
@@ -1,10 +1,44 @@
...
 	github.com/rancher/lasso v0.0.0-20220110204427-db9f3e569574
-	k8s.io/apimachinery v0.20.2
-	k8s.io/client-go v0.20.2
-	sigs.k8s.io/controller-runtime v0.8.2
+	k8s.io/apimachinery v0.23.3
+	k8s.io/client-go v0.23.3
+	sigs.k8s.io/controller-runtime v0.11.0
+)
+
+require (
```
thedadams, line 12: Can these lines be removed to have one `require` directive?

thedadams (changes requested):

controller-runtime/go.mod
maxsokolovsky: I'm having the same thing on my local branch for `rancher/rancher`, and `go mod tidy` forces the two `require` directives for some reason, odd...

kinarashah: ok, so this is a change in behavior with 1.17 
```
Because the number of explicit requirements may be substantially larger in an expanded Go 1.17 go.mod file, the newly-added requirements on indirect dependencies in a go 1.17 module are maintained in a separate require block from the block containing direct dependencies.
``` 
https://go.dev/doc/go1.17#go-command

kinarashah: I think it's also cleaner to look at :)

thedadams (approved):

PR #17, opened 2022-02-09
Add metrics

rmweir:


pkg/metrics/metrics.go
```diff
@@ -0,0 +1,86 @@
...
+	"strconv"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var prometheusMetrics = false
+
+const (
+	lassoSubsystem      = "lasso_controller"
+	controllerNameLabel = "controllerName"
```
ryansann, line 13: I think the common case is snake for label names, any particular reason for the camel case here?

rmweir: Nope, will change.

rmweir: @ryansann updated this label and the ones below it to use snake case.

ryansann (approved):
lgtm

cmurphy (approved):

PR #24, opened 2022-04-08
Add env var for resync interval

rmweir:
**Issue:**
https://github.com/rancher/rancher/issues/37251

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,4 +5,5 @@ go 1.14
 require (
 	k8s.io/apimachinery v0.18.8
 	k8s.io/client-go v0.18.8
+	k8s.io/klog v1.0.0
 )
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -3,13 +3,21 @@ package cache
 import (
 	"context"
 	"fmt"
+	"os"
+	"strconv"
 	"time"
 
 	"github.com/rancher/lasso/pkg/client"
+	"github.com/rancher/lasso/pkg/log"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/tools/cache"
+)
+
+const (
+	// in minutes
+	resyncDefault = 600
 )
 
 type Options struct {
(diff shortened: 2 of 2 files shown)


pkg/cache/cache.go
```diff
@@ -12,6 +14,10 @@ import (
 	"k8s.io/client-go/tools/cache"
 )
 
+const (
+	resyncDefault = 5
```
brandond, line 18: Default interval is 5ns?

Upstream's default resync interval also has a random skew applied; that would probably be a good idea to incorporate here.

brandond (changes requested):

pkg/cache/cache.go
cbron: Do you know where that skew code is Brad ? We had considered doing the same. 
I read this as 5 minutes, but that is also bad, why is it not still 10 hours ? 5 minutes would make everything bunches up and worse.

brandond: For Duration the base integer unit is nanosecond I believe; usually stuff gets parsed from a string or you multiply something by `time.Hour` etc.

I'll see if I can dig up a link to the upstream code.

rmweir: Mb, I had this set to 5 for testing will change. This is in minutes though; I'm multiply with minutes below.

pkg/cache/cache.go
```diff
@@ -53,14 +59,26 @@ func applyDefaultCacheOptions(opts *Options) *Options {
...
 }
 
+func getDefaultResyncInterval() time.Duration {
+	cattleResyncDefaultFromEnv := os.Getenv("CATTLE_RESYNC_DEFAULT")
+	if cattleResyncDefaultFromEnv == "" {
+		return resyncDefault
+	}
+	resyncDefaultFromEnv, err := strconv.Atoi(cattleResyncDefaultFromEnv)
+	if err != nil {
+		return resyncDefault
```
cbron, line 77: Might log error message here that it couldn't read env var, before returning default. Otherwise its a silent error.

cbron (approved):

go.mod
```diff
@@ -5,4 +5,5 @@ go 1.14
 require (
 	k8s.io/apimachinery v0.18.8
 	k8s.io/client-go v0.18.8
+	k8s.io/klog v1.0.0
```
cmurphy, line 8: (nit, feel free to ignore) looks like this is just added by `go mod tidy` and not really related, could be broken into a separate commit

cmurphy (approved):

rmweir:
To test this, setup controllers as seen here https://github.com/rancher/rancher/wiki/How-to-setup-prometheus-dashboard-for-controllers. Then change value and ensure the difference is observed in the handler frequency graph. Spike interval should match the env var setting. The env var being set to "15" should look like this: 
image

The env var being set to 0 should have almost no spikes. Some spikes may be there do to cronjob-like activity but there will be less handlers and the execution number will probably be smaller.

If spikes are difficult to observe then making many resources of a 1-3 types will help. Something like making 1000 secrets.

pkg/cache/cache.go
```diff
@@ -53,14 +61,27 @@ func applyDefaultCacheOptions(opts *Options) *Options {
...
 }
 
+func getDefaultResyncInterval() time.Duration {
+	cattleResyncDefaultFromEnv := os.Getenv("CATTLE_RESYNC_DEFAULT")
+	if cattleResyncDefaultFromEnv == "" {
+		return resyncDefault * time.Minute
+	}
+	resyncDefaultFromEnv, err := strconv.Atoi(cattleResyncDefaultFromEnv)
+	if err != nil {
+		log.Errorf("Lasso: Unable to use resync interval value [%d] from CATTLE_RESYNC_DEFAULT environment variable. Using default [%d].", resyncDefault)
```
rmweir, line 79: This log is missing a value and the first value will be a string not an int.

go.mod
rmweir: Fixed.

cmurphy (approved):

PR #27, opened 2022-04-21
Only handle object if actually updated

rmweir:
**Issue:**
https://github.com/rancher/rancher/issues/37360

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -21,6 +21,10 @@ type Handler interface {
 	OnChange(key string, obj runtime.Object) error
 }
 
+type ResourceVersionGetter interface{
+	GetResourceVersion() string
+}
+
 type HandlerFunc func(key string, obj runtime.Object) error
 
 func (h HandlerFunc) OnChange(key string, obj runtime.Object) error {
@@ -55,7 +59,8 @@ type startKey struct {
 }
 
 type Options struct {
-	RateLimiter workqueue.RateLimiter
+	RateLimiter            workqueue.RateLimiter
+	SyncOnlyChangedObjects bool
 }
 
 func New(name string, informer cache.SharedIndexInformer, startCache func(context.Context) error, handler Handler, opts *Options) Controller {
@@ -72,7 +77,11 @@ func New(name string, informer cache.SharedIndexInformer, startCache func(contex
 	informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: controller.handleObject,
 		UpdateFunc: func(old, new interface{}) {
-			controller.handleObject(new)
+			if !opts.SyncOnlyChangedObjects || old.(ResourceVersionGetter).GetResourceVersion() != new.(ResourceVersionGetter).GetResourceVersion() {
+				// If syncOnlyChangedObjects is disabled, objects will be handled regardless of whether an update actually took place.
+				// Otherwise, objects will only be handled if they have changed
+				controller.handleObject(new)
+			}
 		},
 		DeleteFunc: controller.handleObject,
 	})
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -1,3 +1,7 @@
+/*
+Package controller contains interfaces, structs, and functions that can be used to create shared controllers for
+managing clients, caches, and handlers for multiple types aka GroupVersionKinds.
+*/
 package controller
 
 import (
@@ -21,7 +25,7 @@ type Handler interface {
 	OnChange(key string, obj runtime.Object) error
 }
 
-type ResourceVersionGetter interface{
+type ResourceVersionGetter interface {
 	GetResourceVersion() string
 }
 
--- a/pkg/controller/sharedcontrollerfactory.go
+++ b/pkg/controller/sharedcontrollerfactory.go
@@ -23,11 +23,24 @@ type SharedControllerFactory interface {
 }
 
 type SharedControllerFactoryOptions struct {
+	CacheOptions *cache.SharedCacheFactoryOptions
+
 	DefaultRateLimiter workqueue.RateLimiter
 	DefaultWorkers     int
 
 	KindRateLimiter map[schema.GroupVersionKind]workqueue.RateLimiter
 	KindWorkers     map[schema.GroupVersionKind]int
+
+	// SyncOnlyChangedObjects causes the handle function to only proceed if the object was actually updated.
+	// This is intended to be used by applications with many objects and/or controllers types that have
+	// alternative means of rerunning when necessary. When the informer's resync their cache the update
+	// function is run. If this setting is enabled, when the update handler is triggered the overhead is
+	// reduced but has the tradeoff of not rerunning handlers. Handlers that rely on external objects or
+	// services, or experience a bug might need to rerun despite the respective object not changing. If this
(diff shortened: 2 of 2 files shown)


pkg/controller/controller.go
```diff
@@ -72,7 +81,14 @@ func New(name string, informer cache.SharedIndexInformer, startCache func(contex
 	informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: controller.handleObject,
 		UpdateFunc: func(old, new interface{}) {
-			controller.handleObject(new)
+			if !opts.SkipNoUpdate {
+				// SkipNoUpdate is disabled, objects will occur regardless of whether an update actually took place.
+				controller.handleObject(new)
+			}
+			// SkipNoUpdate is enabled, objects will only be handled if they have changed.
```
cmurphy, line 88: Did you want to add either a return or an else in between these two conditions?

pkg/controller/controller.go
```diff
@@ -1,3 +1,7 @@
+/*
+Package controller contains interfaces, structs, and interfaces that can be used to create shared controllers for
```
cmurphy, line 2:
```suggestion
Package controller contains interfaces, structs, and functions that can be used to create shared controllers for
```

cmurphy (changes requested):

pkg/controller/controller.go
```diff
@@ -72,7 +81,14 @@ func New(name string, informer cache.SharedIndexInformer, startCache func(contex
...
 		UpdateFunc: func(old, new interface{}) {
-			controller.handleObject(new)
+			if !opts.SkipNoUpdate {
+				// SkipNoUpdate is disabled, objects will occur regardless of whether an update actually took place.
+				controller.handleObject(new)
+			}
+			// SkipNoUpdate is enabled, objects will only be handled if they have changed.
+			if old.(ResourceVersionGetter).GetResourceVersion() != new.(ResourceVersionGetter).GetResourceVersion() {
+				controller.handleObject(new)
+			}
```
brandond, line 91:
```suggestion
			// if SkipNoUpdate is enabled, objects will only be handled if they have changed.
			if opts.SkipNoUpdate && old.(ResourceVersionGetter).GetResourceVersion() == new.(ResourceVersionGetter).GetResourceVersion() {
				return
			}
			controller.handleObject(new)
		}
```

pkg/controller/sharedcontrollerfactory.go
```diff
@@ -23,11 +23,22 @@ type SharedControllerFactory interface {
...
 	KindWorkers     map[schema.GroupVersionKind]int
+
+	// SkipNoChangeUpdate causes the handle function to only proceed if the object was actually updated.
+	// This is intended to be used by  applications with many objects and/or controllers types that
+	// have alternative means of rerunning when necessary. When the informer's resync their cache the
+	// update function is run. This reducing the overhead but has the tradeoff of not rerunning handlers
+	// that may rely on external objects or services or experience a bug causing them to not retry even
+	// if it was necessary. The result is that running the handler func on resync will mostly only serve
+	// the purpose of catching missed cache events.
+	SkipNoChangeUpdate bool
```
brandond, line 41: Can we rename this to something like NoopResync or something like that? SkipNoChange is a bit of a confusing double negative - it's not doing something if its not a change?

Making the name consistent across both structs would also be nice.

brandond (changes requested):

pkg/controller/sharedcontrollerfactory.go
rmweir: I feel like the name Noop Resync would be inaccurate since there is an op if the resource versions happen to be different. I also don't think it's a double negative since that is where two negatives achieve the same meaning as a positive which is not what's happening here. The name means that we are skipping the handle function if there is no change in the objects. We can change it to HandleNoChangeUpdate but I'd prefer if the default bool value, false, keeps this feature disabled. Maybe `OnlyUpdateOnObjectChange`?

pkg/controller/controller.go
rmweir: Yes, good catch but ended up combining this anyway per the suggestion below.

pkg/controller/controller.go
```diff
@@ -72,7 +81,11 @@ func New(name string, informer cache.SharedIndexInformer, startCache func(contex
 	informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: controller.handleObject,
 		UpdateFunc: func(old, new interface{}) {
-			controller.handleObject(new)
+			if !opts.SkipNoUpdate || old.(ResourceVersionGetter).GetResourceVersion() != new.(ResourceVersionGetter).GetResourceVersion() {
+				// If SkipNoUpdate is disabled, objects will occur regardless of whether an update actually took place.
```
cmurphy, line 85:
```suggestion
				// If SkipNoUpdate is disabled, objects will be handled regardless of whether an update actually took place.
```

or

```suggestion
				// If SkipNoUpdate is disabled, objects will be updated regardless of whether they actually changed.
```

cmurphy (approved):

brandond (approved):

cmurphy (approved):

pkg/controller/sharedcontrollerfactory.go
```diff
@@ -23,11 +23,22 @@ type SharedControllerFactory interface {
...
 	DefaultRateLimiter workqueue.RateLimiter
 	DefaultWorkers     int
 
 	KindRateLimiter map[schema.GroupVersionKind]workqueue.RateLimiter
 	KindWorkers     map[schema.GroupVersionKind]int
+
+	// SkipNoChangeUpdate causes the handle function to only proceed if the object was actually updated.
+	// This is intended to be used by  applications with many objects and/or controllers types that
+	// have alternative means of rerunning when necessary. When the informer's resync their cache the
+	// update function is run. This reducing the overhead but has the tradeoff of not rerunning handlers
```
cbron, line 37:
```suggestion
	// update function is run. This reduces the overhead, but has the tradeoff of not rerunning handlers
```

pkg/controller/sharedcontrollerfactory.go
```diff
@@ -23,11 +23,22 @@ type SharedControllerFactory interface {
...
+	// that may rely on external objects or services or experience a bug causing them to not retry even
```
cbron, line 38:
```suggestion
	// that may rely on external objects or services, or are experience a bug causing them to not retry even
```

pkg/controller/sharedcontrollerfactory.go
cbron: How about `SyncOnlyChangedObjects`

pkg/controller/sharedcontrollerfactory.go
```diff
@@ -40,6 +51,8 @@ type sharedControllerFactory struct {
 	workers         int
 	kindRateLimiter map[schema.GroupVersionKind]workqueue.RateLimiter
 	kindWorkers     map[schema.GroupVersionKind]int
+
+	SyncOnlyChangedObjects bool
```
brandond, line 55: The rest of these fields don't appear to be exported, is there some reason that this field needs to be? I'm not seeing any use of it outside this package.

rmweir: Fixed.

cbron (approved):

brandond (approved):

PR #29, opened 2022-05-19
Avoid panic if ping is called before the healthcheck is started

ibuildthecloud:


ibuildthecloud:
@kinarashah Any chance this can get merged?

PR #33, opened 2022-10-18
Add useragent

rmweir:
Add the ability to configure useragent
Added the ability to configure a client's user agent. Added wrappers
for controllers, and factories that provide the ability to configure
the user agent of the clients they use.

**Issue:**
https://github.com/rancher/rancher/issues/38775


pkg/client/sharedclientfactory.go
```diff
@@ -188,7 +186,7 @@ func (s *sharedClientFactory) ForResourceKind(gvr schema.GroupVersionResource, k
 	}
 
 	client = NewClient(gvr, kind, namespaced, s.rest, s.timeout)
```
cmurphy, line 188: Would it make sense to just call `.setConfig` from inside `NewClient`?

pkg/client/useragent.go
```diff
@@ -0,0 +1,65 @@
...
+	client := s.SharedClientFactory.ForResourceKind(gvr, kind, namespaced)
+
+	clientWithAgent, err := s.clientWithAgent(client)
+	if err != nil {
+		log.Debugf(fmt.Sprintf("Failed to return client with agent [%s]: %v", s.userAgent, err))
+	}
+	return clientWithAgent
+}
+
+func (s *sharedClientFactoryWithAgent) clientWithAgent(client *Client) (*Client, error) {
```
cmurphy, line 52: Why not use `client.WithAgent`?

rmweir: 🤔 good call, WithAgent properly creates a copy as well. Replaced.

pkg/client/sharedclientfactory.go
rmweir: This was an attempt to maintain compatibility with other consumers of lasso. I figured this approach would be less disruptive since adding it to NewClient would require adding another param which would break anyone that updates lasso.

cmurphy (approved):

pkg/log/log.go
```diff
@@ -10,4 +10,7 @@ var (
 	Errorf = func(message string, obj ...interface{}) {
 		log.Printf("ERROR: "+message+"\n", obj...)
 	}
+	Debugf = func(message string, obj ...interface{}) {
+		log.Printf("DEBUG: "+message+"\n", obj...)
+	}
```
KevinJoiner, line 15: nit:  Shouldn't these just be functions instead of variables pointing to functions?
```suggestion
)
func Debugf(message string, obj ...interface{}) {
	log.Printf("DEBUG: "+message+"\n", obj...)
}
```

pkg/client/client.go
```diff
@@ -38,6 +47,20 @@ func IsNamespaced(gvr schema.GroupVersionResource, mapper meta.RESTMapper) (bool
...
 
+// WithAgent attempts to return a copy of the Client but
+// with a new restClient created with the passed in userAgent.
+func (c *Client) WithAgent(userAgent string) (*Client, error) {
+	client := *c
+	config := c.Config
+	config.UserAgent = userAgent
+	restClient, err := rest.UnversionedRESTClientFor(&config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to created restClient with userAgent [%s]: %v", userAgent, err)
```
KevinJoiner, line 58:
```suggestion
		return nil, fmt.Errorf("failed to created restClient with userAgent [%s]: %w", userAgent, err)
```

KevinJoiner (changes requested):

pkg/client/client_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/client-go/rest"
+)
+
+type userAgentRoundTripper struct {
+	expectedUserAgent string
+	t                 *testing.T
+}
```
mattfarina, line 15: Would it make sense to group the type with its method at the bottom of the file? Just for readability.

pkg/log/log.go
rmweir: I did not write the package and am not sure what the intention was. For now I am conforming to how the other logs forms are written but for the longer term I made this issue for replacing it with an external logging package: https://github.com/rancher/wrangler/issues/244.

pkg/client/client.go
```diff
@@ -83,6 +106,13 @@ func (c *Client) setupCtx(ctx context.Context, minTimeout time.Duration) (contex
 	return context.WithTimeout(ctx, timeout)
 }
 
+func (c *Client) setConfig(config *rest.Config) {
+	if config == nil {
+		return
+	}
```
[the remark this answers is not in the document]
rmweir: I totally agree with this. There is no reason why it wouldn't be passed by value otherwise as well so I changed the function to accept it as a value.

pkg/client/sharedclientfactory.go
```diff
@@ -40,15 +40,12 @@ type sharedClientFactory struct {
 	clients    map[schema.GroupVersionResource]*Client
 	timeout    time.Duration
 	rest       rest.Interface
+	config     *rest.Config
 
 	Mapper meta.RESTMapper
 	Scheme *runtime.Scheme
 }
 
-func NewSharedClientFactoryForConfig(config *rest.Config) (SharedClientFactory, error) {
```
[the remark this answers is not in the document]
rmweir: This was a mistake, ty.

pkg/client/client_test.go
rmweir: I don't have a strong opinion on this. Went ahead and moved it.

pkg/client/client.go
```diff
@@ -83,6 +106,10 @@ func (c *Client) setupCtx(ctx context.Context, minTimeout time.Duration) (contex
 	return context.WithTimeout(ctx, timeout)
 }
 
+func (c *Client) setConfig(config rest.Config) {
+	c.Config = config
```
KevinJoiner, line 110: Since we removed the nil check and now pass by value, we don't need this function we can just set c.Config directly in `ForResourceKind(...)`

KevinJoiner (changes requested):

pkg/client/client.go
```diff
@@ -83,6 +106,10 @@ func (c *Client) setupCtx(ctx context.Context, minTimeout time.Duration) (contex
...
+}
+
```
KevinJoiner, line 112:
```suggestion
```

KevinJoiner (changes requested):

pkg/client/client.go
KevinJoiner: ❓

KevinJoiner: This function is still here

KevinJoiner (changes requested):

mattfarina (approved):
makes sense to me

mattfarina:
@cmurphy do you want to re-review due to changes?

pkg/cache/cache.go
```diff
@@ -1,3 +1,6 @@
+/*
+Package cache provides a cache that maintains a list for a resource type which it updates as it
```
cmurphy, line 2: as it ...what?

sorry, missed this before

mattfarina: @rmweir you've left us hanging in suspense!

rmweir: I think I was mid comment when I realized I didn't end up making any changes to the cache package 😛 . Removed.

cmurphy (approved):

mattfarina (approved):

PR #36, opened 2022-12-05
Add running go tests to drone

rmweir:
**Issue:**
https://github.com/rancher/wrangler/issues/245


cbron (approved):
Would be good to have lint on here as well, but this is ok for now. Why not go generate though ? 
Fwiw we do have a issue tracking these changes: https://github.com/rancher/tasks/issues/201

mattfarina (approved):

PR #37, opened 2023-02-08
Client options

KevinJoiner:
Add lasso options that wrap metav1.Options for Client operations.
Also adds code comments and unit test for Clients.

This PR is needed to add the ability for Rancher to impersonate users for a specific request.


pkg/client/options.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// SetWarningHandler returns a RequestModifier that adds the given warning handler to the request
+func SetWarningHandler(handler rest.WarningHandler) RequestModifier {
+	return func(request *rest.Request) {
+		request.WarningHandler(handler)
+	}
+}
+
+// // Impersonate returns a RequestModifier that will add impersonation headers to the request as the given user and groups.
```
MbolotSuse, line 63: Nit:

We should probably remove all of this commented code

pkg/client/client.go
```diff
@@ -93,36 +102,46 @@ func NewClient(gvr schema.GroupVersionResource, kind string, namespaced bool, cl
...
 		NamespaceIfScoped(namespace, c.Namespaced).
 		Resource(c.resource).
 		Name(name).
-		VersionedParams(&options, metav1.ParameterCodec).
-		Do(ctx).
-		Into(result)
-	return
+		VersionedParams(&opts.GetOptions, metav1.ParameterCodec)
+
+	for _, modifier := range opts.RequestModifiers {
```
MbolotSuse, line 130: Nit:

Might be good to expose a `opts.ApplyModifiers(request)` function so you can avoid doing a for loop each time. It would also allow you to keep the current structure of chained function calls.

pkg/client/client.go
```diff
@@ -132,132 +151,202 @@ func (c *Client) List(ctx context.Context, namespace string, result runtime.Obje
...
+			modifier(req)
+		}
+	}
+
+	return req.Do(ctx).Error()
 }
 
-func (c *Client) DeleteCollection(ctx context.Context, namespace string, opts metav1.DeleteOptions, listOpts metav1.ListOptions) error {
-	ctx, cancel := c.setupCtx(ctx, 0)
+// DeleteCollection will attempt to delete all resource the given namespace (if client.Namespaced is set to true).
```
MbolotSuse, line 295: Nit:
Might be good here to indicate the order that the modifiers are applied. In theory they could overwrite one another so it would be good to define the behavior.

pkg/client/client.go
```diff
@@ -132,132 +151,202 @@ func (c *Client) List(ctx context.Context, namespace string, result runtime.Obje
...
 		Resource(c.resource).
-		VersionedParams(&listOpts, metav1.ParameterCodec).
+		VersionedParams(&listOpts.ListOptions, metav1.ParameterCodec).
 		Timeout(timeout).
-		Body(&opts).
-		Do(ctx).
-		Error()
+		Body(&opts.DeleteOptions)
+
+	for _, modifier := range opts.RequestModifiers {
```
MbolotSuse, line 311: I would recommend that we apply the delete options second. I would expect that we would want those to take precedence in a conflict between the deleteOpts and the listOpts.

pkg/client/client.go
```diff
@@ -132,132 +151,202 @@ func (c *Client) List(ctx context.Context, namespace string, result runtime.Obje
...
+	}
+
+	return req.Do(ctx).Into(result)
 }
 
-func (c *Client) Watch(ctx context.Context, namespace string, opts metav1.ListOptions) (watch.Interface, error) {
+// Watch will attempt to start a watch request with the kube-apiserver for resources in the given namespace (if client.Namespaced is set to true).
+// Results will be streamed too the returned watch.Interface.
+func (c *Client) Watch(ctx context.Context, namespace string, opts ListOptions) (watch.Interface, error) {
 	var timeout time.Duration
```
MbolotSuse, line 171: Nit: 

I'm not sure if this is different because this is a watch which is long-lived, but other functions call c.setupCtx() on the context, so it might be good to have here too.

MbolotSuse (commented):
Couple of questions on the dev-ergonomics of the current approach as well as a couple of nitpicks.

pkg/client/client.go
KevinJoiner: You are correct we do not want to add a timeout on the context because the watch command is an HTTP stream that is long-lived.

pkg/client/client.go
KevinJoiner: Yes I added a function comment as well to address the caller about this.

pkg/client/client.go
KevinJoiner: I added an `apply` function though this does not allow us to chain function calls since we can not add methods to the `rest.Request` struct.

pkg/client/options.go
```diff
@@ -0,0 +1,118 @@
...
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/transport"
+)
+
+// RequestModifier is a function used to modify a request before it is sent.
+type RequestModifier func(*rest.Request)
+
+// Options holds options specific to lasso Clients.
+type Options struct {
+	RequestModifiers []RequestModifier
```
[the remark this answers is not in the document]
KevinJoiner: I updated this so we only allow specific options to be set.

pkg/client/options.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+
+		if 'a' <= b && b <= 'z' || '0' <= b && b <= '9' {
+			buf.WriteByte(b)
+			continue
+		}
+
+		// Kubernetes requires lower-case characters
+		if 'A' <= b && b <= 'Z' {
+			b += 'a' - 'A'
```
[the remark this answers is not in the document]
KevinJoiner: I updated the function to use the `unicode` package purely for clarity. Technically headers only support ASCII which is only the first 128 characters in the Unicode character set.

pkg/client/options.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+		req.SetHeader(transport.ImpersonateGroupHeader, config.Groups...)
+		for k, v := range config.Extra {
+			req.SetHeader(transport.ImpersonateUserExtraHeaderPrefix+sanitizeHeaderKey(k), v...)
+
+		}
+	}
+}
+
+func sanitizeHeaderKey(s string) string {
```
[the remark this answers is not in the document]
KevinJoiner: The docs state the header should be lowercase the client-go implementation is either wrong or they are okay with letting the users send bad requests.
https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation
> Impersonate-Extra-( extra name ): A dynamic header used to associate extra fields with the user. Optional. Requires "Impersonate-User". In order to be preserved consistently, ( extra name ) must be lower-case, and any characters which aren't [legal in HTTP header labels](https://tools.ietf.org/html/rfc7230#section-3.2.6) MUST be utf8 and [percent-encoded](https://tools.ietf.org/html/rfc3986#section-2.1).

I also decided to use a case statement over initializing a map to better emulate the standard library escaping code. https://cs.opensource.google/go/go/+/refs/tags/go1.20.1:src/net/url/url.go;drc=e8fbad5de87f34d2e7632f94cac418c7436174ce;l=101

pkg/client/options.go
```diff
@@ -0,0 +1,118 @@
...
+// 			request.SetHeader("Impersonate-User", username)
+// 		}
+// 		if len(groups) != 0 {
+// 			request.SetHeader("Impersonate-Group", groups...)
+// 		}
+// 	}
+// }
+
+// Impersonate returns a RequestModifier that will add impersonation headers to the request as the given user and groups.
+func Impersonate(config rest.ImpersonationConfig) RequestModifier {
```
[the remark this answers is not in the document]
KevinJoiner: True, K8s could change the values in the future, but the headers are documented in their API thus they will have to follow the proper deprecation cycle which should make this easier to maintain. But that is still a valid concern.

Client-go does not provide any tools for doing this at the moment. the RoundTripper code you linked is a [http.RoundTripper](https://pkg.go.dev/net/http#RoundTripper) which has the function signature of `RoundTrip(*http.Request) (*http.Response, error)`. Client-go does not expose the underlying `http.Request` that is being used or give the ability to add RoundTrippers to a `rest.Request`, which is what we have.

PR #38, opened 2023-02-25
Add to README

rmweir:


README.md
```diff
@@ -1,32 +1,120 @@
...
+Wraps Kubernetes client. Contains GVK fields for managing the client in a parent struct such as the sharedClientyFactory.
+
+### Lasso Cache
+The lasso Cache is a light wrapper around around a [SharedIndexInformer from client-go](https://pkg.go.dev/k8s.io/client-go/tools/cache#SharedIndexInformer).
+
+### Lasso Shared Cache Factory
+The shared cache factory stores the Caches and allows them to be managed via GVK structs. Once the Caches have been created once they can be reused.
+
+### Lasso Shared Client Factory
+The Shared Client Factory exists so that after a client for a GVK has been created once it can be reused.
```
cmurphy, line 92:
```suggestion
The Shared Client Factory exists so that a client for a GVK can be reused after it has been created.
```

README.md
```diff
@@ -1,32 +1,120 @@
...
+
+### Lasso Shared Controller Factory
+The SharedControllerFactory ties everything above together. It allows for a single entry point to access clients, caches, and handlers- all of which it manages for you. It is possible to mainly interact with the shared controller factory. It will create all the underlying controllers, clients, and caches you need. It will ensure there are not duplicates of these objects. It will prevent race conditions.
+
+Once you use ForObject, ForKind, ForResource, or ForResourceKind, the SharedControllerFactory will do the following:
+![SharedControllerFactory Diagram](docs/assets/SharedControllerFactory.png)
+
+### Deferred Controller Function
+
+The deferredController is a function that once executed will created and return a controller. It will create the controller using a cache from the SharedCacheFactory. The deferredController function's purpose is to hold off on populating the cache until absolutely necessary. This ensures only Kubernetes resources that have registered handlers, been enqueued, or have had their controller's cache requested, are being stored in memory.
```
cmurphy, line 102:
```suggestion
The deferredController is a function that, once executed, will created and return a controller. It will create the controller using a cache from the SharedCacheFactory. The deferredController function's purpose is to hold off on populating the cache until absolutely necessary. This ensures only Kubernetes resources that have registered handlers, been enqueued, or have had their controller's cache requested, are being stored in memory.
```

README.md
```diff
@@ -1,32 +1,120 @@
...
+
+When you get a sharedController from the SharedControllerFactory, it won't have an underlying Controller instantiated yet. It will instead called the deferredController method to create one once one of the mentioned operations are performed.
+
+# How to use Lasso
+Below are the steps for a straight forward way to register handlers and run them in lasso:
```
cmurphy, line 107:
```suggestion
Below are the steps for a straightforward way to register handlers and run them in lasso:
```

README.md
```diff
@@ -1,32 +1,120 @@
...
+1. Create rest config pointing to your target cluster. A common way of doing this will be show in example code below.
```
cmurphy, line 108:
```suggestion
1. Create rest config pointing to your target cluster. A common way of doing this will be shown in example code below.
```

README.md
```diff
@@ -38,24 +126,31 @@ import (
 	"k8s.io/client-go/rest"
 )
 
-func MyController(ctx context.Context, config *rest.Config) error {
+func main() {
+	config, err := clientcmd.BuildConfigFromFlags("", os.Getenv("KUBECONFIG_PATH"))
```
cmurphy, line 130: It's usually just 
```suggestion
	config, err := clientcmd.BuildConfigFromFlags("", os.Getenv("KUBECONFIG"))
```

README.md
```diff
@@ -38,24 +126,31 @@ import (
...
+	// ForObject, ForKind, ForResource, and ForResourceKind can all be used to retrieve the controller.
+	// They all accept different parameters but the first three all eventually call ForResourceKind.
+	// Refer to [How to use Lasso](#how-to-use-lasso) for more information on how ForResourceKind works.
 	sharedController, err := controllerFactory.ForObject(&appsv1.Deployment{})
 	if err != nil {
-		return err
+		panic(err)
 	}
 
 	// Register a handler on a shared controller. If you a dedicated queue then use controller.New()
```
cmurphy, line 152:
```suggestion
	// Register a handler on a shared controller. If you need a dedicated queue then use controller.New()
```

README.md
```diff
@@ -74,7 +169,17 @@ func MyController(ctx context.Context, config *rest.Config) error {
 		return result, err
 	}))
 
-	return controllerFactory.Start(ctx, 5)
+	err = controllerFactory.Start(context.TODO(), 5)
```
cmurphy, line 172: Could add a comment here about what `Start` is actually starting and what the number means

README.md
```diff
@@ -1,32 +1,120 @@
...
+### Lasso Controller
+Lasso standardizes what a controller is with its controller package. Lasso introduces factories for managing caches, controllers, and clients within a project. This is imperative for rancher which manages many instances of many types.
+
+The lasso controller uses the client-go work-queue. Lasso registers an event handler that processes the work-queue. The work-queue oriented handler makes it so only one event handler needs to be added for each Kubernetes type. Now registering a handler just adds it to the type's controller's list of handlers. Enqueueing an objects adds it to the work-queue. Each object in the work-queue will then be processed by workers running in go routines. The controller's event handler will re-enqueue any object that causes one or more of the controller's handlers to return an error that is not of type `ErrIgnore`. This work-queue can also have objects enqueued from other sources. This functionality is exposed through the controller's Enqueue function. This transforms controllers from only being reactive to Kubernetes resources and cache re-syncs, to also being invokable by code.
+
+### Lasso Shared Controller
```
cmurphy, line 77: Why is it "shared"?

README.md
```diff
@@ -1,32 +1,120 @@
...
+The sharedController type embeds the Controller interface to expose it's controller's functionality. The sharedController also exposes the RegisterHandler function along with the Client function. RegisterHandler adds the passed handler to the controller's handler list and client returns a client for the sharedController corresponding resource type; sharedController has a GVK field (Group Version Kind) which indicates what type it is for.
+
+The sharedController can create its underlying controller. It will only do so once a handler is registered to the sharedController.
+
+### Lasso Client
+Wraps Kubernetes client. Contains GVK fields for managing the client in a parent struct such as the sharedClientyFactory.
```
cmurphy, line 83:
```suggestion
Wraps Kubernetes client. Contains GVK fields for managing the client in a parent struct such as the sharedClientFactory.
```

What's the value add for using the lasso client vs the k8s client?

README.md
```diff
@@ -1,32 +1,120 @@
...
+The lasso Cache is a light wrapper around around a [SharedIndexInformer from client-go](https://pkg.go.dev/k8s.io/client-go/tools/cache#SharedIndexInformer).
+
+### Lasso Shared Cache Factory
+The shared cache factory stores the Caches and allows them to be managed via GVK structs. Once the Caches have been created once they can be reused.
+
+### Lasso Shared Client Factory
+The Shared Client Factory exists so that after a client for a GVK has been created once it can be reused.
+
+### Lasso Shared Controller Factory
+The SharedControllerFactory ties everything above together. It allows for a single entry point to access clients, caches, and handlers- all of which it manages for you. It is possible to mainly interact with the shared controller factory. It will create all the underlying controllers, clients, and caches you need. It will ensure there are not duplicates of these objects. It will prevent race conditions.
```
cmurphy, line 95: > It will ensure there are not duplicates of these objects

It might be helpful to show an example of what this means, like if I call `controllerFactory.ForObject` twice I won't get a duplicate controller?

cmurphy (commented):
Awesome work :+1:

README.md
```diff
@@ -1,32 +1,120 @@
...
+
+# Lasso's Value Proposition
+### Lasso Controller
+Lasso standardizes what a controller is with its controller package. Lasso introduces factories for managing caches, controllers, and clients within a project. This is imperative for rancher which manages many instances of many types.
+
+The lasso controller uses the client-go work-queue. Lasso registers an event handler that processes the work-queue. The work-queue oriented handler makes it so only one event handler needs to be added for each Kubernetes type. Now registering a handler just adds it to the type's controller's list of handlers. Enqueueing an objects adds it to the work-queue. Each object in the work-queue will then be processed by workers running in go routines. The controller's event handler will re-enqueue any object that causes one or more of the controller's handlers to return an error that is not of type `ErrIgnore`. This work-queue can also have objects enqueued from other sources. This functionality is exposed through the controller's Enqueue function. This transforms controllers from only being reactive to Kubernetes resources and cache re-syncs, to also being invokable by code.
```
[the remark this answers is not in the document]
rmweir: It looks wrong but I believing enqueueing is the proper spelling.

README.md
rmweir: Included this in the step but will comment here as well.

README.md
rmweir: I think it mostly exists to facilitate shareClientFactory by having a GVK field. It also does have some additional options that are not exposed by the Kubernetes client, such as WithAgent and soon impersonation but this is secondary to its intent. I can add a note though.

README.md
rmweir: I think it mostly exists to facilitate shareClientFactory by having a GVK field. It also does have some additional options that are not exposed by the Kubernetes client, such as WithAgent and soon impersonation but this is secondary to its intent. I can add a note though.

README.md
rmweir: I was trying to illustrate these relationships with this [diagram](https://github.com/rancher/lasso/pull/38/files#diff-37acf97508e174947934159a6676cf2cdc0200fce039ea6bda1325e710aa8e1a) below. I thought it would be sufficient between this paragraph and the diagram, but is there something specific that is being lost?

README.md
rmweir: I think the sharing really starts at the more primitive Lasso Controller. The use of work-queue allows one set of Event Handlers to be shared for example. Then the Lasso Shared Controller adds on to this by holding a reusable client for the GVK and adding a way to easily register handlers to the Controller underneath with RegisterHandler.

rmweir:
@cmurphy have addressed most of your feedback and otherwise replied to anything else. I have left my update in a separate commit in case that helps with reviewing. Will squash when ready for approval.

README.md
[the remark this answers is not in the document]
cmurphy: I guess it's both, my bad https://en.wiktionary.org/wiki/enqueuing

README.md
cmurphy: Can you mention that in this description? It's a question I've wondered about when reading the code ("...should I use SharedController or just Controller? I guess I'll just copy what I see without understanding it...") so I think someone else might have this question too. The diagram makes it clearer but just a sentence here saying it can be called again wouldn't hurt.

README.md
cmurphy: Yes that makes it much clearer, sorry for missing that

README.md
rmweir: In most cases, you would use SharedControllerFactory, which will give you a SharedController, which creates a Controller under the hood. I can add to this portion.

README.md
```diff
@@ -1,32 +1,122 @@
...
-Generic form of the generated clientsets.
+If you fall into the category of "Someone who is not new to Kubernetes controllers" you may skip to [Lasso's Value Proposition](#lassos-value-proposition).
 
-#### Package `controller`
+If you are new to Kubernetes controllers it is recommended you do not skip the any sections.
 
-The `Controller` is a simple approach that wraps the standard pattern of worker goroutines and workqueue.  The package
-depends on the `cache` and `client` removing the need for generated clientsets, informers, and listers.
+# Kubernetes Controllers
+_Note:_ This will be a very brief crash course. We encourage you to look up other resources regarding the topic, there are many!
```
moio, line 39: Optional: offer few links to high-quality material.

Rationale: beginners struggle when there is abundance of low quality material.

Personally https://github.com/aiyengar2/k8s-docs/blob/main/docs/controllers/00_introduction.md and the https://web.microsoftstream.com/video/967a4311-fb11-4637-90c9-a4be4c180dad series of videos helped - but I was not starting completely from scratch.

README.md
```diff
@@ -1,32 +1,122 @@
...
+
+In order to achieve the above goal, a few sub-goals need to be met:
+* the controller needs to know about resource type events
+* the controller needs to be passed relevant details regarding the event to a handler
+* (optional) the controller __may__ want to interact with Kubernetes resource types
+
+There are many frameworks to help achieve the above goals. Lasso is one of them. What most of these frameworks have in common, including lasso, is that they use functionality from client-go to achieve these goals. You could achieve these goals by using just client-go. For the remainder of this section, we will fly through how a controller is put together just using client-go concepts.
+
+An informer is used to achieve the above goals needed for a Kubernetes Controller. An informer possesses an event handler and a set of indexers. The underlying structs are different but a pseudo struct is shown below to illustrate this:
```
moio, line 53: Before introducing how an informer is composed, it would help highlighting its role/function/goal.

Actually, since a lot of concepts are introduced, it would help adding definitions for each new jargon term, with examples eg.:

A **resource** is anything Kubernetes manages - such as pods, ingresses, jobs and so on.

An **event** is a change to a resource - such as an addition, an update or a deletion.

An **informer** is a Go object that observes a set of resources through the Kubernetes API and reacts to events pertaining to them. It also keeps a local cache of its objects, to avoid to load the API excessively. You could have your own informer to react to changes in basic Kubernetes resources, or your own.

An **indexer** is a component inside of informers. Its job is to maintain copies of resources temporarily - for example as caches or work queues. As the name implies, indexers keep indices on their resources for fast access.

...and so on

README.md
```diff
@@ -1,32 +1,122 @@
...
+
+Usually an application would want multiple controllers. However, it may not be efficient for every controller to have its own informer and therefore its own indexer. The SharedIndexerInformer from client-go can be used to add multiple controllers that all use the same indexers.
+
+There is a great visual of how a Kubernetes controller might work found in the [kubernetes/sample-controller repo](https://github.com/kubernetes/sample-controller/blob/master/docs/images/client-go-controller-interaction.jpeg).
+
+__Note:__ Informers have been simplified for the sake of clarity. If you are interested in exactly how they work it is recommended you read through the [client-go code](https://github.com/kubernetes/client-go).
+
+# Lasso's Value Proposition
+### Lasso Controller
+Lasso standardizes what a controller is with its controller package. Lasso introduces factories for managing caches, controllers, and clients within a project. This is imperative for rancher which manages many instances of many types.
```
moio, line 75: I think it remains important to highlight that, in Kubernetes jargon, a controller is a pattern, while in Lasso it is an object.

The main pain point Lasso strives to solve is DX in writing controllers: make it as easy as implementing few interfaces, like normal practice in Golang.

README.md
```diff
@@ -1,32 +1,122 @@
...
+### Lasso Shared Controller
+The sharedController type embeds the Controller interface to expose its controller's functionality. The sharedController also exposes the RegisterHandler function along with the Client function. RegisterHandler adds the passed handler to the controller's handler list and Client returns a client for the sharedController corresponding resource type; sharedController has a GVK field (Group Version Kind) which indicates what type it is for.
+
+The sharedController can create its underlying controller. It will only do so once a handler is registered to the sharedController.
+
+### Lasso Client
+Wraps Kubernetes client. Contains GVK fields for managing the client in a parent struct such as the sharedClientFactory. Exposes some additional functionality like configuring the User-Agent header for the underlying REST client.
+
+### Lasso Cache
+The lasso Cache is a light wrapper around a [SharedIndexInformer from client-go](https://pkg.go.dev/k8s.io/client-go/tools/cache#SharedIndexInformer).
```
moio, line 88: Why should I use it instead of SharedIndexInformer?

moio (commented):
Left some totally optional comments - intention being to be even more friendly to absolute beginners 😇

README.md
rmweir: The second link is not open to everyone. I can link the sample controller project instead of that one.

README.md
rmweir: I feel like indexers are further explained here https://github.com/rancher/lasso/pull/38/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R63 and informers are better explained here https://github.com/rancher/lasso/pull/38/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R65. The linked image should also help clarify here https://github.com/rancher/lasso/pull/38/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R69.

README.md
rmweir: I went ahead and updated the opening definition of controller to clarify that it is lasso which uses it to refer to its controller structure(s).

README.md
rmweir: Slightly edited [this comment](https://github.com/rancher/lasso/pull/38/files#r1126898313) and included it in the docs.

README.md
rmweir: Added.

README.md
rmweir: Otherwise, I want to keep this doc more on the lean side. I want to give just enough info if you haven't been exposed to controllers but I don't think this is the place for a general k8s 101.

rmweir: Added a note here: https://github.com/rancher/lasso/pull/38/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R38.

moio: OK, so, I did not write that comment to suggest broadening the scope of the content, but rather to give an example on how adding a bit more structure to the prose could help making it easier to scan for keywords.

A beginner would read the paragraph and extract all the important information. Then go forward with the reading (and perhaps the coding), inevitably forget part of that and at some later point ask "what was an indexer again?" - having a succint, isolated paragraph with **indexer** in bold somewhere helps more than plain prose in answering that question.

That said I completely trust you in knowing readers better and picking the most appropriate final form 👍

moio (approved):
Great work

README.md
```diff
@@ -1,32 +1,128 @@
...
+* (optional) the controller __may__ want to interact with Kubernetes resource types
+
+There are many frameworks to help achieve the above goals. Lasso is one of them. What most of these frameworks have in common, including lasso, is that they use functionality from client-go to achieve these goals. You could achieve these goals by using just client-go. For the remainder of this section, we will fly through how a controller is put together just using client-go concepts.
+
+An informer is used to achieve the above goals needed for a Kubernetes Controller. An informer possesses an event handler and a set of indexers. The underlying structs are different but a pseudo struct is shown below to illustrate this:
+
+```
+Informer
+	EventHandler
+	Indexers
```
KevinJoiner, line 60: I feel like the role of Indexers can get confusing without including the Store at a high level. We do not have to go into the different Store implementations but simply use Stores to help outline the relationship between Informers and Indexers. Such as, an Informer passes objects through indexers to be placed in a Store, and Indexers create indexes on those objects for efficient reads from the store.
```suggestion
Informer
	EventHandler
	Indexers
	Store
```

README.md
```diff
@@ -1,32 +1,128 @@
...
+```
+
+The EventHandler can have different logic for create, update, and delete events. The event handler is where the controller comes in. You use it to configure the create, update, and delete event handlers.
+
+The indexers are a type of computational cache. They are used to map a key to an object, given a function, then the result is saved. Here it is easiest to think of them simply- they receive some metadata such as the name and namespace of the an object and return that object.
+
+The informer first populates the indexer by listing it from a client. Then the informer watches the Kubernetes resource it is assigned to. When an event comes through, it uses that event to update its indexer (by remapping the object associated with the event) and to execute its event handlers.
```
KevinJoiner, line 67:
```suggestion
The informer first populates the indexers and the store by listing all resources returned from a client. Then the informer watches the Kubernetes resource it is assigned to. When an event comes through, it uses that event to update its indexer and execute its event handlers. It does this by remapping the object associated with the event. 
```

KevinJoiner (changes requested):

README.md
```diff
@@ -1,32 +1,128 @@
...
+
+```
+Informer
+	EventHandler
+	Indexers
+```
+
+The EventHandler can have different logic for create, update, and delete events. The event handler is where the controller comes in. You use it to configure the create, update, and delete event handlers.
+
+The indexers are a type of computational cache. They are used to map a key to an object, given a function, then the result is saved. Here it is easiest to think of them simply- they receive some metadata such as the name and namespace of the an object and return that object.
```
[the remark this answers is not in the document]
moio: >  I wouldn't want people to think each indexer has its own cache

Hmm, it is my understanding that `Indexer`s _are_ caches actually - indeed the `Indexer` interface extends `Store`. Now `Store`s (and, by extension, `Indexer`s) can in principle be used for anything - but in practice in client-go they are always used as caches of k8s resource objects (with various usage patterns and deletion policies).

One thing that confused me a lot is that `Indexer`s are not `Indexers`! `Indexers` (with the plural in the interface name!) is, indeed, a completely different concept - a map from `string` to `IndexFunc` - the latter being a function from `any` to a string slice.

Please see:
https://github.com/kubernetes/client-go/blob/1517ffb8d37c99e6a3a2842bcdee0aa271f0332b/tools/cache/index.go#L35-L36
vs
https://github.com/kubernetes/client-go/blob/master/tools/cache/index.go#L97-L98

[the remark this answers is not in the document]
KevinJoiner: > Hmm, it is my understanding that `Indexer`s are caches actually

I agree `Indexer`s non-plural is essentially a cache. 
But the above struct was using plural `Indexers` which is not a cache.

``` 
Informer
	EventHandler
	Indexers
```
We are probably better off explaining that there are two concepts `Indexers`  and `Indexer` and distinguishing them separately to avoid conflation of the two. Especially because anyone using this package will most likely be working with the`Indexers` interface

moio:
@rmweir I think the current text is good, it's an improvement on the earlier revision, and at this point so much better than the original that it should be merged.

Of course refinements by anyone are always possible via PR - I just wouldn't wait for perfection while we have a substantial improvement already 👍 👍 👍

KevinJoiner (approved):
I agree with @moio, I think this captures a lot of good information, that was previously undocumented.

PR #51, opened 2023-08-17
updated k8s-deps to 0.27.4 and controller runtime to 0.15.0

vardhaman22:
related issue: https://github.com/rancher/rancher/issues/41840

for updating controller-runtime changes from #48 are added.


jiaqiluo (changes requested):
We need to bump dependencies in `controller-runtime/go.mod`

kinarashah (changes requested):
Since rancher will be updated to use capi v1.5.0, we'd also need to update corresponding controller-runtime version in lasso. https://github.com/kubernetes-sigs/cluster-api/blob/v1.5.0/go.mod#L46 

https://github.com/rancher/lasso/pull/48 has partial update to v0.14.0 for reference

jiaqiluo (approved):
LGTM

kinarashah (approved):

PR #64, opened 2024-04-16
Use local replace for lasso instead of version number

JonCrowther:
The controller-runtime module uses a lasso version, but that version changes every time we update lasso resulting in an endless loop of updating. Instead, we make use of the "local" version.

[Similar to how we handle rancher/pkg/apis](https://github.com/rancher/rancher/blob/c9be13b09329bbee60a5f6419d500198f83c44d1/go.mod#L12)


tomleb (approved):
This week I tried to find where this package was being used and I couldn't find any place.

According to sourcegraph, it only appears in archived repos: https://sourcegraph.com/search?q=context%3Aglobal+repo%3A%5Egithub%5C.com%2Francher%2F*+%22github.com%2Francher%2Flasso%2Fcontroller-runtime%22&patternType=keyword&sm=0.

JonCrowther:
controller-runtime is removed in https://github.com/rancher/lasso/pull/66

PR #65, opened 2024-04-17
SQLite backed cache

rmweir:
**Issue:**
https://github.com/rancher/rancher/issues/45635
Refers to the [Informer-based, SQLite-backed Steve caching ("Vai")](https://confluence.suse.com/pages/viewpage.action?pageId=1359086083) RFC

**Reviewing:**
I recommend starting with the pkg/cache/sql/README.md. If lasso/controller understanding is limited then read the main README.md first. Once you understand of what informer/indexer/indexer's store generally are and how they work, then move on to the those components of the on-disk cache. A lot of this PR is tests- starting with non-testing files at first should simplify review.

**Testing:**
In addition to passing the unit tests provided in this PR, a version of rancher using this PR has been tested with the integration tests located [here](https://github.com/rmweir/rancher/blob/vai-v2.9/tests/v2/integration/steveapi/steve_api_test.go).


pkg/cache/sql/Readme.md
```diff
@@ -0,0 +1,167 @@
...
+
+### ListOptions Behavior
+Defaults:
+* Sort.PrimaryField: `metadata.namespace`
+* Sort.SecondaryField: `metadata.name`
+* Sort.PrimaryOrder: `ASC` (ascending)
+* Sort.SecondaryOrder: `ASC` (ascending)
+* All filters have partial matching set to false by default
+
+The are some uncommon ways someone could use ListOptions where it would be difficult to predict what the result would be.
```
ericpromislow, line 157: Should be `There are`

pkg/cache/sql/Readme.md
```diff
@@ -0,0 +1,167 @@
...
+		i, err := informerFactory.InformerFor(fields, client, gvk)
+		if err != nil {
+			panic(err)
+		}
+
+		// continueToken will just be an offset that can be used in Resume on a subsequent request to continue
+		// to next page
+		list, continueToken, err := i.ListByOptions(apiOp.Context(), opts, partitions, namespace)
+		if err != nil {
+			return nil, "", err
```
ericpromislow, line 91: I know it's an example, but still should be `panic(err)`

pkg/cache/sql/Readme.md
```diff
@@ -0,0 +1,167 @@
...
+The are some uncommon ways someone could use ListOptions where it would be difficult to predict what the result would be.
+Below is a non-exhaustive list of some of these cases and what the behavior is:
+* Setting Pagination.Page but not Pagination.PageSize will cause Page to be ignored
+* Setting Sort.SecondaryField only will sort as though it was Sort.PrimaryField. Sort.SecondaryOrder will still be applied
+and Sort.PrimaryOrder will be ignored
+
+### Troubleshooting SQLite
+A useful tool for troubleshooting the database files is the sqlite command line tool. Another useful tool is the goland
+sqlite plugin. Both of these tools can be used with the database files. If running a database in-memory, there is no option
+for navigating records. In-memory databases are used for the non-object databse when `CATTLE_ENCRYPT_CACHE_ALL` is set to
```
ericpromislow, line 166: `s/non-object databse/non-object database/`

pkg/cache/sql/Readme.md
```diff
@@ -0,0 +1,167 @@
...
+
+### List Options
+ListOptions includes the following:
+* Match filters for indexed fields. Filters are for specifying the value a given field in an object should be in order to
+be included in the list. Filters can be set to equals or not equals. Filters can be set to look for partial matches or
+exact (strict) matches. Filters can be OR'd and AND'd with one another. Filters only work on fields that have been indexed.
+* Primary field and secondary field sorting order. Can choose up to two fields to sort on. Sort order can be ascending
+or descending. Default sorting is to sort on metadata.namespace in ascending first and then sort on metadata.name.
+* Page size to specify how many items to include in a response.
+* Page number to specify offset. For example, a page size of 50 and a page number of 2, will return items starting at
```
ericpromislow, line 34: I assume the page-number is 1-based?

pkg/cache/sql/Readme.md
```diff
@@ -0,0 +1,167 @@
...
+```sqlite
+JOIN db2."secrets_fields" f ON o.key = f.key
+```
+
+### SQLite Driver
+There are multiple SQLite drivers that this package could have used. One of the most, if not the most, popular SQLite golang
+drivers is [mattn/go-sqlite3](https://github.com/mattn/go-sqlite3). This driver is not being used because it requires enabling
+the cgo option when compiling and at the moment lasso's main consumer, rancher, does not compile with cgo. We did not want
+the SQL informer to be the sole driver in switching to using cgo. Instead, modernc's driver which is in pure golang. Side-by-side
+comparisons can be found indicating the cgo version is, as expect, more performant. If in the future it is deemed worthwhile
```
ericpromislow, line 120: `s/as expect/as expected/`

pkg/cache/sql/attachdriver/driver.go
```diff
@@ -0,0 +1,68 @@
...
+	d := &AttachDriver{
+		path: path,
+		d:    &sqlite.Driver{},
+	}
+	sql.Register("attach_sqlite", d)
+}
+
+// Open opens a connection from the underlying driver and
+// attaches a database as db2 using the underlying path.
+func (d AttachDriver) Open(name string) (driver.Conn, error) {
```
ericpromislow, line 33: I would prefer calling the `AttachDriver` instance `ad` instead of `d` to avoid the `d.d` stutter later down..

pkg/cache/sql/attachdriver/driver.go
```diff
@@ -0,0 +1,68 @@
...
+		return nil, err
+	}
+
+	q := fmt.Sprintf(`ATTACH DATABASE '%s' AS db2;`, d.path)
+	stmt, err := c.Prepare(q)
+	if err != nil {
+		return nil, closeConnOnError(c, err)
+	}
+
+	_, err = stmt.Exec(nil)
```
ericpromislow, line 45: The doc-comment on `Stmt.Exec` says:
```
	// Deprecated: Drivers should implement StmtExecContext instead (or additionally).
	Exec(args []Value) (Result, error)
```

Maybe plan for a separate PR once this PR is merged?

pkg/cache/sql/attachdriver/driver.go
```diff
@@ -0,0 +1,68 @@
...
+	if err != nil {
+		stmt.Close()
```
ericpromislow, line 47: looks like you can do `defer stmt.Close()` after `stmt.Exec()`

pkg/cache/sql/attachdriver/driver.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// closeConnOnError closes the sql.Rows object and wraps errors if needed
+func closeConnOnError(c driver.Conn, err error) error {
+	if err == nil {
+		return nil
+	}
+	ce := c.Close()
+	if ce != nil {
+		if err == nil {
```
ericpromislow, line 61: This condition is impossible because of line 56

pkg/cache/sql/db/client.go
```diff
@@ -0,0 +1,321 @@
...
+func (c *Client) QueryForRows(ctx context.Context, stmt transaction.Stmt, params ...any) (*sql.Rows, error) {
+	c.connLock.RLock()
+	defer c.connLock.RUnlock()
+	var rows *sql.Rows
+	var err error
+
+	err = wait.ExponentialBackoff(backoffRetry, func() (bool, error) {
+		rows, err = stmt.QueryContext(ctx, params...)
+		if err != nil {
+			sqlErr, ok := err.(*sqlite.Error)
```
ericpromislow, line 125: I'm not sure this works because the type of `sqlite.Error` is:
```
type Error struct {
	msg  string
	code int
}
``` 
so it's not a wrapped `error`.

pkg/cache/sql/db/client.go
```diff
@@ -0,0 +1,321 @@
...
+	StmtExec(stmt transaction.Stmt, args ...any) error
+	Exec(stmt string, args ...any) error
+	Commit() error
+	Stmt(stmt *sql.Stmt) transaction.Stmt
+	Cancel() error
+}
+
+// Encryptor encrypts a slice of bytes. It is expected to use data-encryption-key (DEK) which it used to encrypt the slice.
+// It should use a key-encryption-key (KEK) to then encrypt the DEK. The encrypted data, nonce used to encrypt the data,
+// encrypted dek, nonce used to decrypt the dek should be returned, in that order. On failure an error should by returned.
```
ericpromislow, line 72: `s/should by/should be/`

pkg/cache/sql/db/client.go
```diff
@@ -0,0 +1,321 @@
...
+// data-encryption-key (DEK), nonce used to encrypt the DEK. The KEK is not provided. Decryptor is expected to use its
+// receiver to decrypt the DEK using its KEK.
+type Decryptor interface {
+	Decrypt([]byte, []byte, []byte, []byte) ([]byte, error)
+}
+
+var backoffRetry = wait.Backoff{Duration: 50 * time.Millisecond, Factor: 2, Steps: 10}
+
+// NewClient returns a Client. If the given connection is nil then a default one will be created.
+func NewClient(c Connection, encryptor Encryptor, decrypter Decryptor) (*Client, error) {
```
ericpromislow, line 87: This might be deliberate, but I wince because it's spelled `decryptor`. The diff in case is fine

pkg/cache/sql/db/client.go
```diff
@@ -0,0 +1,321 @@
...
+func (c *Client) ReadStrings(rows Rows) ([]string, error) {
+	c.connLock.RLock()
+	defer c.connLock.RUnlock()
+
+	var result []string
+	for rows.Next() {
+		var key string
+		err := rows.Scan(&key)
+		if err != nil {
+			ce := rows.Close()
```
ericpromislow, line 188: Think you can do `return nil, closeRowsOnError(rows, err)` here

And on that subject, I would  have `closeRowsOnError` return `nil, ERROR` so you can just do `return closeRowsOnError(rows, err)`

pkg/cache/sql/db/client.go
```diff
@@ -0,0 +1,321 @@
...
+}
+
+func (c *Client) decryptScan(rows Rows, shouldDecrypt bool) ([]byte, error) {
+	var data, dataNonce, eDEK, dekNonce sql.RawBytes
+	err := rows.Scan(&data, &dataNonce, &eDEK, &dekNonce)
+	if err != nil {
+		return nil, err
+	}
+	if c.decryptor != nil && shouldDecrypt {
+		data, err := c.decryptor.Decrypt(data, dataNonce, eDEK, dekNonce)
```
ericpromislow, line 233: I'm not a fan of shadowed non-error variables. Could you make this `data` something like `decryptedData` ?

pkg/cache/sql/db/client_test.go
```diff
@@ -0,0 +1,515 @@
...
+		client := SetupClient(t, c, e, d)
+		txC := NewMockTXClient(gomock.NewController(t))
+		sqlStmt := &sql.Stmt{}
+		stmt := NewMockStmt(gomock.NewController(t))
+		testObjBytes := toBytes(testObject)
+		testByteValue := []byte("something")
+		e.EXPECT().Encrypt(testObjBytes).Return(testByteValue, testByteValue, testByteValue, testByteValue, nil)
+		txC.EXPECT().Stmt(sqlStmt).Return(stmt)
+		txC.EXPECT().StmtExec(stmt, "somekey", testByteValue, testByteValue, testByteValue, testByteValue).Return(nil)
+		err := client.Upsert(txC, sqlStmt, "somekey", testObject, true)
```
ericpromislow, line 361: Got this error:
```
pkg/cache/sql/db/client_test.go:361:24: cannot use txC (variable of type *MockTXClient) as TXClient value in argument to client.Upsert: *MockTXClient does not implement TXClient (missing method Cancel)
```

pkg/cache/sql/db/client_test.go
```diff
@@ -0,0 +1,515 @@
...
+		e := SetupMockEncryptor(t)
+		d := SetupMockDecryptor(t)
+		expectedClient := Client{
+			conn:      c,
+			encryptor: e,
+			decryptor: d,
+		}
+		client, err := NewClient(c, e, d)
+		assert.Nil(t, err)
+		assert.Equal(t, expectedClient, *client)
```
ericpromislow, line 44: My IDE complains that `expectedClient` doesn't have a `connLock  sync.RWMutex` field

pkg/cache/sql/db/transaction/transaction.go
```diff
@@ -0,0 +1,94 @@
+/*
+Package transaction provides a client for a live transactions, and interfaces for some relevant sql types. The transaction client automatically performs rollbacks  on failures.
```
ericpromislow, line 2: `s/transactions/transaction`

pkg/cache/sql/encryption/encrypt.go
```diff
@@ -0,0 +1,182 @@
+/*
+package encryption provide a struct, Manager, that generates a key-encryption-key, generates and rotates a
```
ericpromislow, line 2: `s/provide/provides/`

pkg/cache/sql/encryption/encrypt.go
```diff
@@ -0,0 +1,182 @@
...
+data-encryption-key, and provides functions for encryption and decryption. The key-encryption-key is used so the client
+can store an encrypted version of their data key. This key-hierarchy enable the rotation of data-encryption-keys which
```
ericpromislow, line 4: `s/enable/enables/`

pkg/cache/sql/encryption/encrypt.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+func (m *Manager) isWriteCounterOverMax() bool {
+	return m.writesCounter > maxWriteCount
+}
+
+// fetchUpToDateDataKey returns the current data key if writeCounter has not exceeded maxWriteCount. If it has then the
+// data key is rotated. Before exiting the writesCounter is incremented by 1. This means that one use of the key has
+// been accounted for. This is done so encryption does not block other processes from encrypting. The consequence is a
+// key should only be used once and fetchUpToDateDataKey should be used to ensure an up-to-date key for any proceeding
```
ericpromislow, line 143: I think you mean `s/proceeding/subsequent/` ?

pkg/cache/sql/informer/factory/informer_factory.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+
+		return i, nil
+	}
+
+	return result, nil
+}
+
+// Reset closes the stopCh which stops any running informers, assigns a new stopCh, resets the GVK-informer cache, and resets
+// the database connection which wipes any current sqlite databse at the default location.
```
ericpromislow, line 117: `s/databse/database/`

pkg/cache/sql/Readme.md
rmweir: If you mean do page numbers start at one, yes. I added some clarification

pkg/cache/sql/attachdriver/driver.go
rmweir: I think the reason is that I want to close the stmt before closing the connection which happen in closeOnError(c, err) since they are coupled. If I defer then then it will try to close stmt on a close connection. Added a comment.

pkg/cache/sql/attachdriver/driver.go
rmweir: Fixed and changed how I'm wrapping error below.

pkg/cache/sql/db/client.go
rmweir: It implements error and is returned the sqlite driver returns them.

pkg/cache/sql/db/client.go
rmweir: Replace occurences of these with closeRowsOnError. I can have it return nil because I call it in `ReadObjects` which has a different return type. Updated function comment for closeRowsOnError as well, name was wrong.

pkg/cache/sql/db/client.go
rmweir: Yep, meant to get them all the same my bad.

pkg/cache/sql/db/client_test.go
rmweir: Made some recent changes, I plan on getting back to these tests. Will push other parts first then come back to this.

pkg/cache/sql/db/client_test.go
rmweir: I believe its complaining that they're both not pointers and will end up with the exact same locks. I'm not sure why I compared their values. Maybe I mistakenly though it would compare their addresses but it doesn't. Should be gone now.

pkg/cache/sql/attachdriver/driver.go
ericpromislow: ok

pkg/cache/sql/db/client.go
ericpromislow: yup

ericpromislow (dismissed):
Only remaining issue is that `go test ./...` fails with type-related errors in function calls

MbolotSuse (commented):
Couple of comments, mostly on the informer factory and the informer

pkg/cache/sql/informer/informer.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	loi, err := NewListOptionIndexer(fields, s, namespaced)
+	if err != nil {
+		return nil, err
+	}
+
+	// HACK: replace the default informer's indexer with the SQL based one
+	UnsafeSet(sii, "indexer", loi)
+
+	return &Informer{
```
MbolotSuse, line 68: From what I can tell here, what we are doing here is the following:
- We are constructing a `sharedIndexInformer`. 
- Our `Informer` struct is implementing the `SharedIndexInformer` interface.
- We are overriding the internal `indexer` field of the struct to be our new indexer, which also uses a SQL store. 
- We are implementing this `ListByOptions` function, which allow us to be more creative in how the SQL store is accessed.

I've got a couple of overall questions on this:
- SharedIndexInformer is an interface. Rather than trying to override an unexported field on the concrete type, could we instead implement our own version of this interface?
  - Internally, this could construct a `SharedInformer` and use `AddEventHandler` to allow us to index these types without implementing the run loop.
    - This does have a memory cost since the original informer would still store objects.
    - We could vendor the run loop at the moment to avoid this.
- The listOptions/Partition logic seems like something that belongs to a consuming library. It's using the indexer for enhanced querying, but this isn't logic that helps to run a controller, and it doesn't conform to the indexer interface. Can you elaborate on how this fits within Lasso? 
  - To go a bit further, it looks like the SharedInformer would allow us to work with a store that doesn't have indexers. If we are going to use something other than an indexer to query values, it seems like it would make sense to use that interface rather than the SharedIndexerInformer one. Though I might be missing what we get out of the indexer in this model.

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -0,0 +1,455 @@
...
+
+	// 2- Sorting: ORDER BY clauses (from lo.Sort)
+	orderByClauses := []string{}
+	if len(lo.Sort.PrimaryField) > 0 {
+		columnName := toColumnName(lo.Sort.PrimaryField)
+		direction := "ASC"
+		if lo.Sort.PrimaryOrder == DESC {
+			direction = "DESC"
+		}
+		orderByClauses = append(orderByClauses, fmt.Sprintf(`f."%s" %s`, columnName, direction))
```
MbolotSuse, line 262: The column name seems to be end-user provided input. This seems like a case for injection to me. Can you give this a look and let me know what you think?

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -0,0 +1,455 @@
...
+	}
+	if len(lo.Sort.SecondaryField) > 0 {
+		columnName := toColumnName(lo.Sort.SecondaryField)
+		direction := "ASC"
+		if lo.Sort.SecondaryOrder == DESC {
+			direction = "DESC"
+		}
+		orderByClauses = append(orderByClauses, fmt.Sprintf(`f."%s" %s`, columnName, direction))
```
MbolotSuse, line 270: The column name seems to be end-user provided input. This seems like a case for injection to me. Can you give this a look and let me know what you think?

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -0,0 +1,455 @@
...
+	var orWhereClause string
+	var params []any
+
+	for index, filter := range orFilters.Filters {
+		opString := "LIKE"
+		if filter.Op == NotEq {
+			opString = "NOT LIKE"
+		}
+		columnName := toColumnName(filter.Field)
+		orWhereClause += fmt.Sprintf(`f."%s" %s ?`, columnName, opString)
```
MbolotSuse, line 359: The column name seems to be end-user provided input. This seems like a case for injection to me. Can you give this a look and let me know what you think?

pkg/cache/sql/partition/partition.go
```diff
@@ -0,0 +1,17 @@
...
+package partition
+
+import (
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// Partition represents filtering of a request's results by namespace or a list of resource names
+type Partition struct {
+	Namespace   string
+	All         bool
```
MbolotSuse, line 14: It would be worth commenting each of these items. From what I was reading, it looks like if you don't set All you need to set at least one Name, but that's not documented here.

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -0,0 +1,455 @@
...
+				params = append(params, partition.Namespace)
+			}
+
+			// optionally filter by names
+			if !partition.All {
+				names := partition.Names
+
+				if len(names) == 0 {
+					// degenerate case, there will be no results
+					singlePartitionClauses = append(singlePartitionClauses, "FALSE")
```
MbolotSuse, line 217: Nit: Rather than appending a clause which will return no results, could we instead just return some sort of default here? I'm open to other ideas though,either case is probably ok.

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -0,0 +1,455 @@
...
+	return orWhereClause, params
+}
+
+// toColumnName returns the column name corresponding to a field expressed as string slice
+func toColumnName(s []string) string {
+	return strings.Join(s, ".")
+}
+
+// getField extracts the value of a field expressed as a string path from an unstructured object
+func getField(a any, field string) (any, error) {
```
MbolotSuse, line 379: How does this differ from the existing logic in wrangler: https://github.com/rancher/wrangler/blob/652f4c907eb5181b45f1ab7a51d2e4cd390911c1/pkg/data/values.go#L49? Is there any overlap that we can take advantage of so that we can avoid duplicating logic (either here or in wrangler)?

rmweir: They're different in what they provide. Field in getField supports the full path and in doing so provides the field parsing. It is meant to accept fields as they appear when it comes to slices/maps, like `spec.someMap["cattle.io/somekey"]` where it needs to distinguish between a `.` that separates fields and ones that are part of keys. It also provides the option to return a field of all elements of a slice, delimited with `|`. This is useful for filtering container images in pods, for example. They probably could at least use each other but I think they're different enough with this one being tailored to this particular use case that I don't think its needed.

pkg/cache/sql/informer/informer.go
rmweir: Can you clarify what you mean when you say we could construct a SharedInformer internally? Are you proposing we still use the sharedIndexInformer type but wrap it and expose a run that just uses event handler?

We provide informer backed caches in lasso. The store that we are extending now provides new functionality which forms queries based on the the options struct. In that sense I think it's the best place to provide it. It also supports the concept of partitions as priority zero filters for the store. I'm not entirely sure how any of this does not fit in lasso. Sure there are other ways we can move the structs among our 3 to 4 repos but I'm not sure it gets any more logical.

pkg/cache/sql/informer/factory/informer_factory.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	dbClient, err := db2.NewClient(nil, m, m)
+	if err != nil {
+		return nil, err
+	}
+	return &InformerFactory{
+		wg:          wait.Group{},
+		stopCh:      make(chan struct{}),
+		cache:       map[schema.GroupVersionKind]*informer.Informer{},
+		encryptAll:  os.Getenv(EncryptAllEnvVar) == "true",
```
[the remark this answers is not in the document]
rmweir: The agreement with security was for secrets and tokens to be encrypted by default with the option to toggle encryption for all. Pinging @macedogm for confirmation. This is also in doc and was signed off by security.

pkg/cache/sql/informer/factory/informer_factory.go
```diff
@@ -0,0 +1,134 @@
...
+		dbClient:    dbClient,
+		newInformer: informer.NewInformer,
+	}, nil
+}
+
+// InformerFor returns an informer for given GVK, using sql store indexed with fields, using the specified client
+func (f *InformerFactory) InformerFor(fields [][]string, client dynamic.ResourceInterface, gvk schema.GroupVersionKind, namespaced bool) (*informer.Informer, error) {
```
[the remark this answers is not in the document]
rmweir: At the moment indexes are per gvk and hard-coded or driven off the discovery api/crds in steve. When crds change the cache is simply restarted. Indexes are always type wide. The flow is basically hey does this informer exist?  if it doesn't then make it and index with these fields, if it does then my ignore fields bc they are deterministic and won't have anything new anyway.

pkg/cache/sql/informer/factory/informer_factory.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+const (
+	EncryptAllEnvVar = "CATTLE_ENCRYPT_CACHE_ALL"
+)
+
+func init() {
+	indexedFieldDBPath := db2.OnDiskInformerIndexedFieldDBPath
+	if os.Getenv(EncryptAllEnvVar) == "true" {
+		indexedFieldDBPath = ":memory:" // indexedFieldDBPath = "file::memory:?cache=shared"
```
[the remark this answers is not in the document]
rmweir: Yes, since we can't encrypt these because then you would not be able to query off of them. So far the performance has seemed good enough where I am tempted to have it always set to in-memory and just have the option for non memory for dev debugging.

pkg/cache/sql/partition/partition.go
```diff
@@ -0,0 +1,17 @@
...
+	Passthrough bool
+	Names       sets.String
```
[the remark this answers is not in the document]
rmweir: I think I attempted to change this in the past but for some reason thought I couldn't. Maybe I thought I'd have to change it everywhere including in steve access set logic, not sure. Anyways I have switched it now.

go.mod
```diff
@@ -2,21 +2,33 @@ module github.com/rancher/lasso
...
 	github.com/prometheus/client_golang v1.16.0
+	github.com/sirupsen/logrus v1.9.3
 	github.com/stretchr/testify v1.9.0
 	k8s.io/api v0.28.6
 	k8s.io/apimachinery v0.28.6
-	k8s.io/client-go v0.28.6
+	k8s.io/client-go v12.0.0+incompatible
 	k8s.io/klog v1.0.0
+	k8s.io/utils v0.0.0-20230406110748-d93618cff8a2
+	modernc.org/sqlite v1.27.0
```
macedogm, line 22: Is it possible to use the current latest version instead [v1.29.10](https://gitlab.com/cznic/sqlite/-/tags/v1.29.10)?

pkg/cache/sql/informer/factory/informer_factory.go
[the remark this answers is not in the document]
macedogm: > The agreement with security was for secrets and tokens to be encrypted by default with the option to toggle encryption for all.

Yes, that is the agreement, by default only sensitive fields are encrypted, for security and efficiency reasons, but the option to encrypt everything is provided.

If I understood correctly, [these](https://github.com/rancher/lasso/pull/65/files#diff-30ea1a028c2e55261cf0c7fd16267f209c53187a2fa4372afc2580001b9a668bR51-R56) are the default encrypted fields, right?

Question that I don't recollected the answer, where will these settings be documented?

pkg/cache/sql/informer/listoption_indexer.go
MbolotSuse: > Field in getField supports the full path and in doing so provides the field parsing.

I'm not sure what you mean here, the wrangler function does this as well AFAICT.

> It is meant to accept fields as they appear when it comes to slices/maps, like spec.someMap["cattle.io/somekey"] where it needs to distinguish between a . that separates fields and ones that are part of keys.

Thanks for elaborating, that helps.

Overall, I don't think that the use case really differs between these functions. They are both for retrieving field from unstructured objects. I think we want to limit the amount of functions that provide this utility for debugging reasons. That being said, I think that we could use a sort of "util" library for things like this, so I won't hold things up on this point.

pkg/cache/sql/store/store.go
```diff
@@ -0,0 +1,370 @@
...
+	Upsert(tx db.TXClient, stmt *sql.Stmt, key string, obj any, shouldEncrypt bool) error
+	CloseStmt(closable db.Closable) error
+	// TxModeClient() DBClient
+}
+
+var backoffRetry = wait.Backoff{Duration: 50 * time.Millisecond, Factor: 2, Steps: 10}
+
+// NewStore creates a SQLite-backed cache.Store for objects of the given example type
+func NewStore(example any, keyFunc cache.KeyFunc, c DBClient, shouldEncrypt bool, name string) (*Store, error) {
+	s := &Store{
```
pjbgf, line 74: Some workflows result in `panic`, so please add `nil` checks for `keyFunc` and `c`. We probably should add a `nil` check for the return of `reflect.TypeOf(example)` as well.

pkg/cache/sql/store/store.go
```diff
@@ -0,0 +1,370 @@
...
+
+type DBClient interface {
+	Begin() (db.TXClient, error)
+	Prepare(stmt string) *sql.Stmt
+	QueryForRows(ctx context.Context, stmt transaction.Stmt, params ...any) (*sql.Rows, error)
+	ReadObjects(rows db.Rows, typ reflect.Type, shouldDecrypt bool) ([]any, error)
+	ReadStrings(rows db.Rows) ([]string, error)
+	Upsert(tx db.TXClient, stmt *sql.Stmt, key string, obj any, shouldEncrypt bool) error
+	CloseStmt(closable db.Closable) error
+	// TxModeClient() DBClient
```
pjbgf, line 67:
```suggestion
```

pkg/cache/sql/store/store.go
```diff
@@ -0,0 +1,370 @@
...
+	return bb
+}
+
+// fromBytes decodes an object from a byte slice
+func (s *Store) fromBytes(buf sql.RawBytes) (reflect.Value, error) {
+	dec := gob.NewDecoder(bytes.NewReader(buf))
+	singleResult := reflect.New(s.typ)
+	err := dec.DecodeValue(singleResult)
+	return singleResult, err
+}
```
pjbgf, line 323: This is not called anywhere here, removing all the tests still pass. Can we just remove it from the codebase?

```suggestion
```

pkg/cache/sql/informer/indexer.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+// Test that Indexer implements cache.Indexervar _ cache.Indexer = (*Indexer)(nil)
+
+// NewIndexer returns a cache.Indexer backed by SQLite for objects of the given example type
+func NewIndexer(indexers cache.Indexers, s Store) (*Indexer, error) {
+	// sanity checks first
+	for key := range indexers {
+		if strings.Contains(key, `"`) {
+			panic("Quote characters (\") in indexer names are not supported")
```
pjbgf, line 86: Can't we just return an error instead?

pkg/cache/sql/store/store.go
```diff
@@ -0,0 +1,370 @@
...
+	CloseStmt(closable db.Closable) error
+	// TxModeClient() DBClient
+}
+
+var backoffRetry = wait.Backoff{Duration: 50 * time.Millisecond, Factor: 2, Steps: 10}
+
+// NewStore creates a SQLite-backed cache.Store for objects of the given example type
+func NewStore(example any, keyFunc cache.KeyFunc, c DBClient, shouldEncrypt bool, name string) (*Store, error) {
+	s := &Store{
+		name:          sanitize(name),
```
pjbgf, line 75: `NewIndexer` panics when name contains `"`, here we just remove them and carry on.
We could align behaviours and return an error here for invalid table names.

pkg/cache/sql/attachdriver/driver.go
```diff
@@ -0,0 +1,62 @@
...
+
+// Open opens a connection from the underlying driver and
+// attaches a database as db2 using the underlying path.
+func (ad AttachDriver) Open(name string) (driver.Conn, error) {
+	c, err := ad.d.Open(name)
+	if err != nil {
+		return nil, err
+	}
+
+	q := fmt.Sprintf(`ATTACH DATABASE '%s' AS db2;`, ad.path)
```
pjbgf, line 38: This is prone to SQL injection. If path is `file:/tmp/sqlite' as DB2;ATTACH DATABASE "/tmp/target" AS tg; CREATE TABLE tg.abc (data text); INSERT INTO tg.abc (data) VALUES ("123456");--`, an arbitrary sqlite db (`/tmp/target`) will be created with whatever values defined within the `path`.

Let's add some validation and make this value quite strict.

pkg/cache/sql/attachdriver/driver.go
rmweir: @ericpromislow the used driver has not implemented `StmtExecContext` yet.

pkg/cache/sql/informer/indexer.go
```diff
@@ -0,0 +1,258 @@
...
+	ReadStrings(rows db.Rows) ([]string, error)
+	Prepare(stmt string) *sql.Stmt
+	CloseStmt(stmt db.Closable) error
+}
+
+// Test that Indexer implements cache.Indexervar _ cache.Indexer = (*Indexer)(nil)
+
+// NewIndexer returns a cache.Indexer backed by SQLite for objects of the given example type
+func NewIndexer(indexers cache.Indexers, s Store) (*Indexer, error) {
+	// sanity checks first
```
pjbgf, line 83: `Store` is an interface, the use of `s.GetName()` below assumes that it depends on a concrete implementation which already sanitised its contents for SQL injection, which may not always hold true. This results in a few places using `s.GetName()` in this file being vulnerable to SQL injection.

pjbgf (changes requested):

pkg/cache/sql/attachdriver/driver.go
moio: Right: https://gitlab.com/cznic/sqlite/-/blob/master/sqlite.go?ref_type=heads#L1700

pkg/cache/sql/Readme.md
```diff
@@ -0,0 +1,167 @@
...
+
+### Encryption Defaults
+By default only specified types are encrypted. These types are hard-coded and defined by defaultEncryptedResourceTypes
+in `pkg/cache/sql/informer/factory/informer_factory.go`. To enabled encryption for all types, set the ENV variable
+`CATTLE_ENCRYPT_CACHE_ALL` to "true".
+
+The key size used is 256 bits. Data-encryption-keys are stored in the object table and are rotated every 150,000 writes.
+
+### Indexed Fields
+Filtering and sorting only work on indexed fields. These fields are defined when using Informer For. Objects will
```
pjbgf, line 144: Maybe?

```suggestion
Filtering and sorting only work on indexed fields. These fields are defined when using `InformerFor`. Objects will
```

pkg/cache/sql/informer/listoption_indexer.go
rmweir: Validated column names and added guidelines to README.md on how to write queries.

pkg/cache/sql/informer/listoption_indexer.go
rmweir: Same as https://github.com/rancher/lasso/pull/65#discussion_r1612435288.

pkg/cache/sql/informer/listoption_indexer.go
rmweir: Same as https://github.com/rancher/lasso/pull/65#discussion_r1612435288.

pkg/cache/sql/Readme.md
rmweir: Fixed and changed to `CacheFor`.

pkg/cache/sql/informer/factory/informer_factory.go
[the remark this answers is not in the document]
rmweir: Tokens are not accessible via the v1 endpoint atm https://github.com/rancher/rancher/blob/release/v2.9/pkg/api/steve/disallow/disallow.go#L39.

pkg/cache/sql/db/client.go
```diff
@@ -0,0 +1,313 @@
...
+	return tx.StmtExec(tx.Stmt(stmt), key, objBytes, dataNonce, encryptedDEK, dekNonce)
+}
+
+// toBytes encodes an object to a byte slice
+func toBytes(obj any) []byte {
+	var buf bytes.Buffer
+	enc := gob.NewEncoder(&buf)
+	err := enc.Encode(obj)
+	if err != nil {
+		panic(errors.Wrap(err, "Error while gobbing object"))
```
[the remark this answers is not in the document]
moio: Hmm, unsure. What should Rancher do with that error? Potentially important data intended to be saved is now lost, because it did not reach the DB.

Reading into gob's source, this method will return an error if a nil pointer is passed, or if a type points to itself in a cycle (eg. `T *T`), or if the type is not gobbable for some other reason. In all those cases the only possibility is a code fix, therefore failing fast seems like the best option to me, and restarting Rancher the least worse option at disposal.

What do you think?

pkg/cache/sql/informer/factory/informer_factory.go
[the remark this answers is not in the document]
moio: I suggest leaving as-is for now, decision can always be reconsidered at a later point.

pkg/cache/sql/informer/factory/informer_factory.go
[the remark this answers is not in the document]
moio: > where will these settings be documented?

It will all go to official docs, likely starting from an experimental paragraph in the 'Tuning and Best Practices for Rancher at Scale' section.

I plan to write that as soon as we have confirmation from QA that no big regressions exist with the option turned off.

pkg/cache/sql/informer/indexer.go
```diff
@@ -0,0 +1,258 @@
...
+	"github.com/rancher/lasso/pkg/cache/sql/db"
+	"github.com/rancher/lasso/pkg/cache/sql/db/transaction"
+	"k8s.io/client-go/tools/cache"
+)
+
+const (
+	selectQueryFmt = `
+			SELECT object FROM "%s"
+				WHERE key IN (
+					SELECT key FROM "%s_indices"
```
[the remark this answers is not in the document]
moio: 🥷 thanks for making me aware of this swift syntax possibility!

pkg/cache/sql/informer/indexer.go
moio: Well there is only one Store implementation, and it actually belongs to this library and package, and it does sanitize the name. The reason for having an interface here is mocking.

Are we really targeting a threat model in which a _programmer_ uses the library in an incorrect way, as opposed to a _user_ trying to inject SQL commands?

If we are, how important is this from your perspective?

pkg/cache/sql/informer/informer.go
moio: About pairing our SQL-based `Store` to a `SharedInformer` instead of `SharedIndexInformer` to trim all `Indexer` related fat we do not actually use: it'd be a great idea!

There's a catch though. The problem is we'd have to replicate some logic from client-go's implementation, because the provided `SharedInformer` implementation is simply using a `SharedIndexInformer` instance implementation with an empty `Indexer` list under the hood 😢:

https://github.com/rancher/client-go/blob/b4d94f01600c2edfda43d82052bb0a14d0bdb120/tools/cache/shared_informer.go#L162-L163

Reimplementing the `SharedInformer` interface, then, would mean IIUC replicating ~600 lines from shared_informer.go plus tests and dependencies (from the definition of `sharedIndexInformer` on).

I do not think we want to go down that particular path (note: implementing something other than `SharedInformer` is different).

pkg/cache/sql/informer/listoption_indexer.go
moio: We have three cases to "encode" into `singlePartitionClauses`:

1. the normal case: one or two clauses to be `AND`ed together as filtering (WHERE) conditions. Currently encoded as a non-empty list of clauses
2. the `partition.All` degenerate case: no filtering conditions. Currently encoded as an empty list of clauses
3. the no-Names degenerate case: no results _from this partition_. Currently encoded as `FALSE`

Note that 3. does not mean we can immediately `return toUnstructuredList([]any{}), nil, nil` - other partitions might have results and all `singlePartitionClauses` have to be `OR`ed together.

We could encode 3. in another variable (eg a bool) to avoid the `OR (FALSE)` "nop" clause - `singlePartitionNoResults` maybe? I was not sure this is actually easier to comprehend.

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -0,0 +1,476 @@
...
+			}
+
+			if len(singlePartitionClauses) > 0 {
+				partitionClauses = append(partitionClauses, strings.Join(singlePartitionClauses, " AND "))
+			}
+		}
+	}
+	if len(partitions) == 0 {
+		// degenerate case, there will be no results
+		whereClauses = append(whereClauses, "FALSE")
```
moio, line 235: @MbolotSuse this case is actually different and `return toUnstructuredList([]any{}), nil, nil` would be valid here, IIUC

pkg/cache/sql/partition/partition.go
moio: Agreed on comments

Setting All to false and Names to empty set would just mean you get no results from this Partition. Not a particularly useful query I agree - still legit

pkg/cache/sql/store/store.go
```diff
@@ -0,0 +1,370 @@
...
+// Test that Store implements cache.Indexer
+var _ cache.Store = (*Store)(nil)
+
+type DBClient interface {
+	Begin() (db.TXClient, error)
+	Prepare(stmt string) *sql.Stmt
+	QueryForRows(ctx context.Context, stmt transaction.Stmt, params ...any) (*sql.Rows, error)
+	ReadObjects(rows db.Rows, typ reflect.Type, shouldDecrypt bool) ([]any, error)
+	ReadStrings(rows db.Rows) ([]string, error)
+	Upsert(tx db.TXClient, stmt *sql.Stmt, key string, obj any, shouldEncrypt bool) error
```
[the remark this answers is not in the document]
moio: IIUC intent is to abstract away something that can store and retrieve objects SQL storage + blob encoding (gob) + optional encryption "layer"

Encryption ends up determining a part of the structure of tables/columns (`object, dek, deknonce, datanonce`) which ends up influencing upsert, select and create table statements.

So DBClient ends up with knowledge about, and DBClient users have to rely on this implicit assumption.

I will think about if and how this could be decoupled better.

pkg/cache/sql/informer/factory/informer_factory.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	attachdriver.Register("file:" + indexedFieldDBPath + "?cache=shared")
+}
+
+// NewInformerFactory returns an informer factory instance
+func NewInformerFactory() (*InformerFactory, error) {
+	m, err := encryption.NewManager()
+	if err != nil {
```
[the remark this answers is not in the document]
rmweir: @MbolotSuse there is at least one type that is always encrypted, Secrets and in the future tokens. I do think the types should be move up to steve and passed into lasso. At that time this can check if there are any types that will need encryption. For the present though there is no use case with some sort of encryption.

pkg/cache/sql/informer/factory/informer_factory.go
```diff
@@ -0,0 +1,134 @@
...
+	}, nil
+}
+
+// InformerFor returns an informer for given GVK, using sql store indexed with fields, using the specified client
+func (f *InformerFactory) InformerFor(fields [][]string, client dynamic.ResourceInterface, gvk schema.GroupVersionKind, namespaced bool) (*informer.Informer, error) {
+	f.informerCreateLock.Lock()
+	defer f.informerCreateLock.Unlock()
+
+	result, ok := f.cache[gvk]
+	if !ok {
```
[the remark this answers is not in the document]
rmweir: I agree, that's better. Fixed.

pkg/cache/sql/informer/factory/informer_factory.go
```diff
@@ -0,0 +1,134 @@
...
+
+	return result, nil
+}
+
+// Reset closes the stopCh which stops any running informers, assigns a new stopCh, resets the GVK-informer cache, and resets
+// the database connection which wipes any current sqlite database at the default location.
+func (f *InformerFactory) Reset() error {
+	if f.dbClient == nil {
+		// nothing to reset
+		return nil
```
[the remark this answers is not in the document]
rmweir: We make the assumption here that if dbClient is not set then nothing else has to be reset. We could reset the map but that would imply the assumption might be wrong in which case we also can't assume the status of stopCh which would be an issue.

pkg/cache/sql/informer/informer.go
```diff
@@ -0,0 +1,82 @@
...
+type ByOptionsLister interface {
+	ListByOptions(ctx context.Context, lo ListOptions, partitions []partition.Partition, namespace string) (*unstructured.UnstructuredList, string, error)
+}
+
+// NewInformer returns a new SQLite-backed Informer for the type specified by schema in unstructured.Unstructured form
+// using the specified client
+func NewInformer(client dynamic.ResourceInterface, fields [][]string, gvk schema.GroupVersionKind, db sqlStore.DBClient, shouldEncrypt bool, namespaced bool) (*Informer, error) {
+	listWatcher := &cache.ListWatch{
+		ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
+			a, err := client.List(context.TODO(), options)
```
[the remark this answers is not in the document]
rmweir: Stopping is handled by the reflector. When the stop channel is closed, the watch interfaces will be closed https://github.com/kubernetes/client-go/blob/master/tools/cache/reflector.go#L677. I can switch to context.Background().

pkg/cache/sql/informer/informer.go
```diff
@@ -0,0 +1,82 @@
...
+			return a, err
+		},
+		WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+			return client.Watch(context.TODO(), options)
```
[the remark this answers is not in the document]
rmweir: See https://github.com/rancher/lasso/pull/65#discussion_r1617988383.

pkg/cache/sql/store/store.go
```diff
@@ -0,0 +1,370 @@
...
+	s.deleteStmt = s.Prepare(fmt.Sprintf(deleteStmtFmt, s.name))
+	s.getStmt = s.Prepare(fmt.Sprintf(getStmtFmt, s.name))
+	s.listStmt = s.Prepare(fmt.Sprintf(listStmtFmt, s.name))
+	s.listKeysStmt = s.Prepare(fmt.Sprintf(listKeysStmtFmt, s.name))
+
+	return s, nil
+}
+
+/* Core methods */
+// TODO: this is where I left off, need to figure out how lock will be handled
```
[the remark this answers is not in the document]
rmweir: Yep, removed.

pkg/cache/sql/store/store.go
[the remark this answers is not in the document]
rmweir: I think there are plenty of different ways the db layer could be designed. To be honest the largest value I was looking to extract was testability, so I needed some  of layer there. It provides a database client, encrypting/decrypting like @moio pointed out, some automatic closing, and a connection reset that also cleans files. I think if desired it could always be changed later. I think a challenge in having a different client for each type will be using the same same connection for all of them. In most cases it's not recommended to have more than 1 or 2 (if you want to separate read and write) since a connection in golang's sql package is a pool of connections that are being coordinated for you.

pkg/cache/sql/partition/partition.go
rmweir: Added comment to function. Yes, as Silvio said that'll work but will result in an empty list.

pkg/cache/sql/db/client.go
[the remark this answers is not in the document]
pjbgf: My concern here was that specially craft input could trigger the panic therefore crashing the entire application, potentially causing a DoS. On the next review I will take a deeper look at this.

pkg/cache/sql/informer/informer.go
tomleb: > Reimplementing the SharedInformer interface, then, would mean IIUC replicating ~600 lines from shared_informer.go plus tests and dependencies (from the definition of sharedIndexInformer on).

Agreed. The way I see it, if we want a `cache.SharedIndexInformer` then we're better off re-using the one from client-go instead of re-implmenting those 600+ lines (via unsafe hack/client-go fork patch). I think the `Store`+`Controller` approach would be useful in the case that we don't want the `cache.SharedIndexInformer`.

pkg/cache/sql/informer/indexer.go
```diff
@@ -0,0 +1,255 @@
...
+	rows, err := i.QueryForRows(context.TODO(), stmt, params...)
+	if err != nil {
+		return nil, err
+	}
+	return i.ReadObjects(rows, i.GetType(), i.GetShouldEncrypt())
+}
+
+// ByIndex returns the stored objects whose set of indexed values
+// for the named index includes the given indexed value
+func (i *Indexer) ByIndex(indexName, indexedValue string) ([]any, error) {
```
tomleb, line 194: Trying to use the ByIndex method here: https://github.com/tomleb/lasso-controller-notes/blob/f4b0ad310f76999edfadaf03c82f5308f6cdf3b6/simple-vai/main.go#L87-L101 and it fails with 

```
[] sql: expected 1 destination arguments in Scan, not 4
```

I have tried using an upstream `cache.SharedIndexInformer` and the same example works fine. Am I missing something in the example that would make it work?

pkg/cache/sql/informer/indexer.go
```diff
@@ -0,0 +1,255 @@
...
+		return nil, err
+	}
+	return i.ReadObjects(rows, i.GetType(), i.GetShouldEncrypt())
+}
+
+// IndexKeys returns a list of the Store keys of the objects whose indexed values in the given index include the given indexed value
+func (i *Indexer) IndexKeys(indexName, indexedValue string) ([]string, error) {
+	i.indexersLock.RLock()
+	defer i.indexersLock.Unlock()
+	indexFunc := i.indexers[indexName]
```
tomleb, line 206:
```suggestion
func (i *Indexer) IndexKeys(indexName, indexedValue string) ([]string, error) {
	i.indexersLock.RLock()
	defer i.indexersLock.RUnlock()
	indexFunc := i.indexers[indexName]
```

Trying to use the IndexKeys method here: https://github.com/tomleb/lasso-controller-notes/blob/f4b0ad310f76999edfadaf03c82f5308f6cdf3b6/simple-vai/main.go#L87-L101 and it panics with 

```
fatal error: sync: Unlock of unlocked RWMutex
```

Fixing the mutex issue, I then get an unexpected result. I get `[]` but I should get all the keys for this indexKey+indexValue pair (which would be all ConfigMaps in this example).

I have tried using an upstream `cache.SharedIndexInformer` and I get the expected result (all configmap keys). Am I missing something in the example that would make it work?

pkg/cache/sql/informer/indexer.go
```diff
@@ -0,0 +1,255 @@
...
+		}
+	}
+	return nil
+}
+
+// add afterDelete
+/* Satisfy cache.Indexer */
+
+// Index returns a list of items that match the given object on the index function
+func (i *Indexer) Index(indexName string, obj any) ([]any, error) {
```
tomleb, line 152: Trying to use the Index method here: https://github.com/tomleb/lasso-controller-notes/blob/f4b0ad310f76999edfadaf03c82f5308f6cdf3b6/simple-vai/main.go#L87-L101 and it fails with 

```
[] sql: expected 1 destination arguments in Scan, not 4
```

I have tried using an upstream `cache.SharedIndexInformer` and the same example works fine. Am I missing something in the example that would make it work?

pkg/cache/sql/informer/indexer.go
```diff
@@ -0,0 +1,255 @@
...
+
+	rows, err := i.QueryForRows(context.TODO(), i.listKeysByIndexStmt, false, indexName, indexedValue)
+	if err != nil {
+		return nil, err
+	}
+	return i.ReadStrings(rows)
+}
+
+// ListIndexFuncValues wraps safeListIndexFuncValues and panics in case of I/O errors
+func (i *Indexer) ListIndexFuncValues(name string) []string {
```
tomleb, line 219: Trying to use the ListIndexFuncValues method here: https://github.com/tomleb/lasso-controller-notes/blob/f4b0ad310f76999edfadaf03c82f5308f6cdf3b6/simple-vai/main.go#L87-L101 and it returns an empty list of keys instead of a list of all configmap keys.

I have tried using an upstream `cache.SharedIndexInformer` and the same example works fine. Am I missing something in the example that would make it work?

tomleb (commented):
I have updated to the latest changes and tested the use of the custom Indexer here:  https://github.com/tomleb/lasso-controller-notes/blob/f4b0ad310f76999edfadaf03c82f5308f6cdf3b6/simple-vai/main.go#L87-L101.

This example file seem to work fine with upstream `cache.SharedIndexInformer`, so I assume it should/we want it to also work with this one.

pkg/cache/sql/informer/indexer.go
tomleb: Might be worth adding _some_ unit tests that would catch at least the locking issue.

pkg/cache/sql/informer/indexer.go
tomleb: > Are we really targeting a threat model in which a programmer uses the library in an incorrect way, as opposed to a user trying to inject SQL commands?

I don't think those two are mutually exclusing. A programmer incorrectly using the library could lead to SQL injection down the line.

> The reason for having an interface here is mocking.

I'm still getting used to mocking for testing so sorry if that's a stupid question but do we need mocks here considering we're using sqlite as the only implementation and that it supports running in memory via `:memory:`?

If so and if we want to avoid programmer errors that lead to SQL injection, could we have this function be internal only and perhaps a less generic `NewIndexer` function for library use that only accepts the sqlite store? This way we can make sure that whoever is using lasso in rancher/k3s/rke2/wtv can only pass in a "safe" implementation.

rmweir: but we are similarly relying on indexed fields https://github.com/rancher/lasso/pull/65/files#diff-910681530ceac7cfe911fb307754542734adee41c8d009d795aabe0d9e3803b2R25 for validation which could also be configured to inject, and a developer could misuse partitions. A developer could expose any factory in lasso and that would also be unsafe. I just think lasso is covered in comparable cases where if you just used it as is it would be insecure to hand to a user. It’s an admin level tool.

I don’t think it hurts to move santize() to the db package, export it, and use it liberally where we perform sql queries. I went ahead and did that. I just don’t think the original logic tracks for me.

@tomleb I think it's better not to incur the overhead of running sqlite as the job of the unit tests are to test the functionality contained in the method. The store is also tested and sqlite is tested by the maintainers. We then have integration tests which test them working together. It also could affect running the tests in parallel.

pkg/cache/sql/informer/factory/informer_factory.go
```diff
@@ -0,0 +1,134 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/tools/cache"
+	"os"
+	"sync"
+)
+
+// InformerFactory builds Informer instances and keeps a cache of instances it created
+type InformerFactory struct {
+	informerCreateLock sync.Mutex
```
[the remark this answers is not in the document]
rmweir: I agree RWMutex should be used where possible, fixed.

pkg/cache/sql/db/client.go
```diff
@@ -0,0 +1,313 @@
...
+	err := os.RemoveAll(InformerObjectCacheDBPath)
+	if err != nil {
+		return err
+	}
+
+	err = os.RemoveAll(OnDiskInformerIndexedFieldDBPath)
+	if err != nil {
+		return err
+	}
+	sqlDB, err := sql.Open("attach_sqlite", "file:"+InformerObjectCacheDBPath+"?mode=rwc&cache=shared&_journal_mode=wal&_synchronous=off&_foreign_keys=on&_busy_timeout=1000000")
```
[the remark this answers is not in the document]
pjbgf: We probably want to log in case the file already exists and their permissions are too loose.

pkg/cache/sql/informer/factory/informer_factory.go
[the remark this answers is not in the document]
MbolotSuse: Resolving since we will always use some form of encryption.

pkg/cache/sql/informer/factory/informer_factory.go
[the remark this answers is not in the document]
MbolotSuse: Given that all of the args are non-exported I think that's fine. I'll rsolve.

pkg/cache/sql/informer/informer.go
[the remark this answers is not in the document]
moio: -> [epic](https://github.com/rancher/rancher/issues/45635)

pkg/cache/sql/informer/factory/informer_factory.go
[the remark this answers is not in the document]
moio: -> [epic](https://github.com/rancher/rancher/issues/45635)

pkg/cache/sql/store/store.go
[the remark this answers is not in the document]
moio: -> [epic](https://github.com/rancher/rancher/issues/45635)

pkg/cache/sql/store/store.go
```diff
@@ -0,0 +1,370 @@
...
+
+	err = s.Upsert(tx, s.upsertStmt, key, obj, s.shouldEncrypt)
+	if err != nil {
+		return err
+	}
+
+	err = s.runAfterUpsert(key, obj, tx)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
moio: I agree, this needs refactoring.

Either the registered function has the explicit duty of rolling back the transaction, or this code should do that.

Currently it's mixed: if the called code does (eg. by using the provided tx.StmtExec func) then it will perform the rollback, if failing for other reasons it won't.

-> [epic](https://github.com/rancher/rancher/issues/45635)

pkg/cache/sql/informer/informer.go
```diff
@@ -0,0 +1,82 @@
...
+		},
+	}
+
+	example := &unstructured.Unstructured{}
+	example.SetGroupVersionKind(gvk)
+
+	// avoids the informer to periodically resync (re-list) its resources
+	// currently it is a work hypothesis that, when interacting with the UI, this should not be needed
+	resyncPeriod := time.Duration(0)
```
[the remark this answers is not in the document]
moio: -> [epic](https://github.com/rancher/rancher/issues/45635)

pkg/cache/sql/informer/listoption_indexer.go
moio: My understanding is that this mechanism is used:

- for `defaultIndexedFields`/`defaultIndexNamespaced`: constants, using only the `.` operator
- for fields computed from schemas (`getFieldsFromSchema` in steve): using only the `.` operator
- for `typeSpecificIndexedFields` in steve: constant, using both the `.` and the `[]` operator depending on the type (only a handful of them are defined as of today)

If all we needed was the `.` operator, then this "mini-language parser" would not be needed, as [][]string suffices (and indeed was implemented that way in an earlier version).

Given we also need the `[]` operator we thus need something more powerful. We could settle on this "mini-language", but from the UI team we got indication that will also not suffice, because of Steve-specific virtual fields like `id` or `metadata.name`. Therefore I believe we will need a more flexible approach passing down a `func` instead of a string in this "mini-language" (which is, by the way, exactly what Indexers do).

Conclusion: I expect this code to be completely changed.

-> [epic](https://github.com/rancher/rancher/issues/45635)

pkg/cache/sql/store/store.go
moio: To me if either `keyFunc` or `c` are nil then it is a programmer error without any reasonable ways to recover, so I am fine panicking.

Let's have a discussion via chat if you have strong perspectives on this one.

pkg/cache/sql/store/store.go
moio: Agreed, will be fixed as part of a second look at SQL injection from the perspective of non-Steve or future users of lasso (Steve currently always passes a Kubernetes type name which can't have those chars in anyway).

-> [epic](https://github.com/rancher/rancher/issues/45635)

pkg/cache/sql/attachdriver/driver.go
moio: Fixed

pkg/cache/sql/db/client.go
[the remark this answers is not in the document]
moio: -> [epic](https://github.com/rancher/rancher/issues/45635)

pkg/cache/sql/informer/indexer.go
moio: -> [epic](https://github.com/rancher/rancher/issues/45635)

pkg/cache/sql/db/transaction/transaction.go
```diff
@@ -0,0 +1,94 @@
...
+// Commit commits the transaction and then unlocks the database.
+func (c *Client) Commit() error {
+	c.closed = true
+	return c.sqlTx.Commit()
+}
+
+// Exec uses the sqlTX Exec() with the given stmt and args. The transaction will be automatically rolled back if Exec()
+// returns an error.
+func (c *Client) Exec(stmt string, args ...any) error {
+	_, err := c.sqlTx.Exec(stmt, args...)
```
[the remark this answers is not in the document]
moio: Agreed we had a problem here.

The only purpose of closed was to avoid a double rollback on Close, but that had a race condition possibility as access to close was not serialized.

Fixed in 12a491e

pkg/cache/sql/db/client.go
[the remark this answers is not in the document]
moio: I do not disagree that may happen (though unlikely, as objects are expected to come from Kubernetes, where such conditions should not be possible).

What I mean is that, even in case it does happen, we would not have a much better alternative to bubble up the error all the way up to the main method and basically restart the application anyway.

I am resolving this for now, let's revisit when the new audit happens.

pkg/cache/sql/db/client.go
```diff
@@ -0,0 +1,312 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	"modernc.org/sqlite"
+	sqlite3 "modernc.org/sqlite/lib"
+)
+
+const (
+	// InformerObjectCacheDBPath is where SQLite's object database file will be stored relative to process running lasso
+	InformerObjectCacheDBPath = "informer_object_cache.db"
+	// OnDiskInformerIndexedFieldDBPath is where SQLite's indexed fields database file will be stored if, env var value
+	// found at EncryptAllEnvVar is "true".
```
tomleb, line 29:
```suggestion
	// OnDiskInformerIndexedFieldDBPath is where SQLite's indexed fields database file will be stored if, env var value
	// found at EncryptAllEnvVar is "false".
```

My understanding is that if the CATTLE_ENCRYPT_CACHE_ALL env var is `true` then the index is in memory instead.

pkg/cache/sql/db/client.go
```diff
@@ -0,0 +1,312 @@
...
+
+	"github.com/pkg/errors"
+	"github.com/rancher/lasso/pkg/cache/sql/attachdriver"
+	"github.com/rancher/lasso/pkg/cache/sql/db/transaction"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"modernc.org/sqlite"
+	sqlite3 "modernc.org/sqlite/lib"
+)
+
+const (
```
tomleb, line 25: Are the two databases split so that we can have the `db2` one in memory if encryption is required?

Can that cause any issues with transaction? I couldn't find if SQLite with WAL journal mode supports transaction with multi-databases.

tomleb (commented):
Had some pending comments I never sent. Nit, some clarifying questions, with the assumption that we want to merge and do some more work after in https://github.com/rancher/rancher/issues/45635.

I'd have like to look at the partition bit but Michael already did a good job reviewing there so I'll look at it post-merge.

pkg/cache/sql/db/client.go
rmweir: @tomleb yes they are split and that is why. The purpose of attach driver is to make sure the second db2 is part of the connection. I haven't read anything that would make me think it would be an issue, especially because we are reusing the connection pool.

pkg/cache/sql/store/store.go
```diff
@@ -0,0 +1,333 @@
...
+
+	txC, err := s.Begin()
+	if err != nil {
+		return err
+	}
+
+	for _, key := range keys {
+		err = txC.StmtExec(txC.Stmt(s.deleteStmt), key)
```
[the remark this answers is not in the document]
rmweir: Regardless of the two functions being called concurrently, the fact we are beginning a transaction which locks the sqlite db file for write should prevent the situation where add's upsert happens in between the delete and upsert of replace. That is if I understand your concern correctly.

pkg/cache/sql/db/client.go
rmweir: Yes, that is the reason for db2 running in memory and the purpose of attach driver is to attach every underlying connection started by golang's sql package.  It sounds like the only scenario in which there can be an issue is if if it is WAL and the program crashes. This case does not apply to us since we do not reuse the database between rancher restarts. See https://www.sqlite.org/lang_attach.html:

> Transactions involving multiple attached databases are atomic, assuming that the main database is not "[:memory:](https://www.sqlite.org/inmemorydb.html)" and the [journal_mode](https://www.sqlite.org/pragma.html#pragma_journal_mode) is not [WAL](https://www.sqlite.org/wal.html). If the main database is ":memory:" or if the journal_mode is WAL, then transactions continue to be atomic within each individual database file. But if the host computer crashes in the middle of a [COMMIT](https://www.sqlite.org/lang_transaction.html) where two or more database files are updated, some of those files might get the changes where others might not.

pkg/cache/sql/informer/indexer.go
rmweir: These were all due to bugs which I have now found and resolved. Should add integration tests for this. It may be difficult to include in a rancher integration test but they can be setup for lasso. Here are the reasons for each failure:
1.) When [adding encryption](https://github.com/rmweir/lasso/blob/vai/pkg/cache/sql/db/client.go#L219) the object scan was adjusted to accept the object columns and the 3 encryption related columns. The [select query](https://github.com/rmweir/lasso/blob/vai/pkg/cache/sql/informer/indexer.go#L18) was not updated to also select these columns so it ended up passing a number of rawbytes pointers that didn't match the number of values queried.
2.) Same as 1, but for the [list by index query](https://github.com/rmweir/lasso/blob/vai/pkg/cache/sql/informer/indexer.go#L35).
3.) Used Unlock() when RUnlock() is needed.
4.) For some reason [false was being passed as a param to the query](https://github.com/rmweir/lasso/blob/vai/pkg/cache/sql/informer/indexer.go#L211).
5.) [Same as 4 when querying for index values](https://github.com/rmweir/lasso/blob/vai/pkg/cache/sql/informer/indexer.go#L229).

moio:
> @moio I had some concern about some of the `cache.Indexer` / `cache.Store` methods panicking (`ListIndexFuncValues`, `List` and `ListKeys`). I understand that those interface don't allow us to return an error, which is a bummer, but I'm wondering if panicking is the best we can do here given that this is a library.
> 
> Have we thought about other ways of handling these failing cases (eg: returning an empty list)?

To be honest, no, but I see your point.

> Is this something we could perhaps revisit in [rancher/rancher#45635](https://github.com/rancher/rancher/issues/45635)? Panicking seems pretty brutal for something that could have been SQLite being busy, as an example.

SQLite being busy is addressed by retrying code already, and the number of retries should be so high for that to signal a real bug rather than a temporary condition (or at least, that was the design intent).

But I agree there might be other causes and panic might be inappropriate. I added a point to the epic issue and welcome discussion on alternatives.

pkg/cache/sql/store/store.go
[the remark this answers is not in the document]
moio: Good catch, solved with fd6f9d0

@rmweir's reasoning is correct in principle, the problem was that code did not make `listKeysStmt` part of the transaction, so it was executed in another transaction.

By making the SELECT query part of the same transaction that runs the DELETE and UPSERT queries, we get the "snapshot isolation" guarantee from SQLite: if another concurrent transaction has committed mid-processing, this transaction fails with a busy error and will need to be retried.

moio:
> Only remaining issue is that `go test ./...` fails with type-related errors in function calls

Tests all pass now, dismissing review.

PR #66, opened 2024-04-19
Remove controller-runtime module

tomleb:
# Problem

This module doesn't seem to be used and is somewhat a burden:

- This module has its own `go.mod` (duh) which means:
  - It needs to be updated every time we update k8s libraries (eg: https://github.com/rancher/lasso/pull/62).
  - We get renovate bot PRs for it
  - It depends on top-level lasso module at a specific commit, meaning it's never in sync (we wanted to address it here: https://github.com/rancher/lasso/pull/64).
- It doesn't have any test, and since we don't know how it's used it's difficult to know if something would break current usage.

It doesn't make sense to keep maintaining this if it's not being used, so this is an attempt to remove it.

# Research

I've manually checked the following repos (and their corresponding branch of rancher 2.9, 2.8, 2.7):
- apiserver
- norman
- steve
- webhook
- remotedialer
- dynamiclistener
- wrangler

Lasso is used by other projects as well (mostly via wrangler), so I also checked the following:
- fleet
- rke2
- rke
- aks-operator
- eks-operator
- gke-operator
- k3s

No usage in any of those repos.

[Using sourcegraph](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/rancher+%22github.com/rancher/lasso/controller-runtime%22&patternType=keyword&sm=0), I don't see anything **still in development** within `github.com/rancher` that is using this package.
[
Including archived repositories in the result](https://sourcegraph.com/search?q=context%3Aglobal+repo%3A%5Egithub%5C.com%2Francher+%22github.com%2Francher%2Flasso%2Fcontroller-runtime%22&patternType=keyword&sm=0&filters=%5B%5B%22utility%22%2C%22Include+archived+repos%22%2C%22archived%3Ayes%22%5D%5D), we can see that https://github.com/rancher/rancher-operator is using it. However this repository has been archived last year (and hasn't been touched in 3 years).

Let's see if any team responds that they need it, otherwise we can remove it.


brandond:
I suspect the idea was to ease migration to/from upstream controller-runtime by adding a translation between lasso caches and controller-runtime caches... but since this is a classic Darren direct-push-to-master commit with no PR or context in the commit message we can only guess.

Kill it with fire!

brandond (approved):

PR #67, opened 2024-04-24
[v1.29] Bump lasso

krunalhinguu:
### Updates

- updated lasso to latest

### Related

- https://github.com/rancher/rancher/issues/43110


MbolotSuse (changes requested):
@krunalhinguu As seen in #66, we actually want to outright remove the `controller-runtime` module. Can you give that a look at integrate with that, or just wait for that PR to go in?

PR #69, opened 2024-04-26
feat(sharedcachefactory): allow checking if a given kind is being cached

aruiz14:
Issue: https://github.com/rancher/rancher/issues/41905

This PR extends the `SharedCacheFactory` interface to allow checking if a specific kind is being cached.
This will allow Norman clients to check this before calling `Controller()`, which implicitly registers the kind in the cache.

Note: With the goal of using the existing mocks, I've copied them to a separate internal package, but couldn't remove the existing one since `clientMutator.go` is within the same package as the interface being mocked.


ericpromislow (approved):
Looks straightforward

PR #80, opened 2024-06-28
Transition to simpler key management

pjbgf:
Move away from key hierarchy keys, into simpler data slicing instead.

The use of key hierarchy, saving the DEKs alongside the data, is very
effective when there is a need for long-term storage. As Lasso's KEK is
not persisted externally, once the instance restarts the encrypted data
can no longer be used as the DEKs can't be decrypted.

Due to that, a more performant approach would be do away with the KEK, and
keep all the DEKs in memory. The stored data now has an uint32 keyID, as opposed
to having an encrypted DEK and its nonce - saving 40 bytes per row.

As a result, the encryption process had a throughput increase of ~260MB/s
and a 22% decrease in allocations. Conversely, the decryption process had
throughput increase of ~400MB/s and a 50% decrease in allocations.

```
goos: linux
goarch: amd64
pkg: github.com/rancher/lasso/pkg/cache/sql/encryption
cpu: 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
                           │ /tmp/before  │             /tmp/after              │
                           │    sec/op    │   sec/op     vs base                │
Encryption/encrypt-1024-16    1.416µ ± 0%   1.013µ ± 1%  -28.46% (p=0.000 n=10)
Encryption/encrypt-4096-16    2.240µ ± 0%   1.853µ ± 0%  -17.32% (p=0.000 n=10)
Encryption/encrypt-8192-16    3.279µ ± 1%   2.844µ ± 0%  -13.28% (p=0.000 n=10)
Decryption/decrypt-1024-16   1044.5n ± 1%   662.8n ± 4%  -36.54% (p=0.000 n=10)
Decryption/decrypt-4096-16    1.910µ ± 1%   1.541µ ± 3%  -19.34% (p=0.000 n=10)
Decryption/decrypt-8192-16    3.101µ ± 3%   2.617µ ± 1%  -15.61% (p=0.000 n=10)
geomean                       2.002µ        1.557µ       -22.21%

                           │ /tmp/before  │              /tmp/after               │
                           │     B/s      │      B/s       vs base                │
Encryption/encrypt-1024-16   689.7Mi ± 0%    964.2Mi ± 1%  +39.79% (p=0.000 n=10)
Encryption/encrypt-4096-16   1.702Gi ± 0%    2.059Gi ± 0%  +20.97% (p=0.000 n=10)
Encryption/encrypt-8192-16   2.327Gi ± 1%    2.683Gi ± 0%  +15.32% (p=0.000 n=10)
Decryption/decrypt-1024-16   934.6Mi ± 1%   1473.4Mi ± 4%  +57.64% (p=0.000 n=10)
Decryption/decrypt-4096-16   1.997Gi ± 1%    2.475Gi ± 3%  +23.97% (p=0.000 n=10)
Decryption/decrypt-8192-16   2.460Gi ± 2%    2.916Gi ± 1%  +18.49% (p=0.000 n=10)
geomean                      1.512Gi         1.944Gi       +28.56%

                           │ /tmp/before  │              /tmp/after              │
                           │     B/op     │     B/op      vs base                │
Encryption/encrypt-1024-16   2.078Ki ± 0%   2.016Ki ± 0%   -3.01% (p=0.000 n=10)
Encryption/encrypt-4096-16   5.703Ki ± 0%   5.641Ki ± 0%   -1.10% (p=0.000 n=10)
Encryption/encrypt-8192-16   10.20Ki ± 0%   10.14Ki ± 0%   -0.61% (p=0.000 n=10)
Decryption/decrypt-1024-16   2.781Ki ± 0%   1.875Ki ± 0%  -32.58% (p=0.000 n=10)
Decryption/decrypt-4096-16   5.781Ki ± 0%   4.875Ki ± 0%  -15.68% (p=0.000 n=10)
Decryption/decrypt-8192-16   9.781Ki ± 0%   8.875Ki ± 0%   -9.27% (p=0.000 n=10)
geomean                      5.166Ki        4.590Ki       -11.16%

                           │ /tmp/before │             /tmp/after             │
                           │  allocs/op  │ allocs/op   vs base                │
Encryption/encrypt-1024-16    9.000 ± 0%   7.000 ± 0%  -22.22% (p=0.000 n=10)
Encryption/encrypt-4096-16    9.000 ± 0%   7.000 ± 0%  -22.22% (p=0.000 n=10)
Encryption/encrypt-8192-16    9.000 ± 0%   7.000 ± 0%  -22.22% (p=0.000 n=10)
Decryption/decrypt-1024-16   12.000 ± 0%   6.000 ± 0%  -50.00% (p=0.000 n=10)
Decryption/decrypt-4096-16   12.000 ± 0%   6.000 ± 0%  -50.00% (p=0.000 n=10)
Decryption/decrypt-8192-16   12.000 ± 0%   6.000 ± 0%  -50.00% (p=0.000 n=10)
geomean                       10.39        6.481       -37.64%
```


pkg/cache/sql/encryption/encrypt.go
```diff
@@ -133,50 +121,56 @@ func createGCMCypher(key []byte) (cipher.AEAD, error) {
...
-// key should only be used once and fetchUpToDateDataKey should be used to ensure an up-to-date key for any subsequent
-// encryption.
-func (m *Manager) fetchUpToDateDataKey() ([]byte, error) {
+// fetchActiveDataKey returns the current data key and its key ID.
+// Each call results in activeKeyCounter being incremented by 1. When the
+// the activeKeyCounter exceeds maxWriteCount, the active data key is
+// rotated - before being returned.
+func (m *Manager) fetchActiveDataKey() ([]byte, uuid.UUID, error) {
 	m.lock.Lock()
-	defer m.lock.Unlock()
```
moio, line 147: What pros and cons do you see with using `defer` here?

pkg/cache/sql/store/store.go
```diff
@@ -17,17 +17,16 @@ import (
...
+	listStmtFmt     = `SELECT object, objectnonce, dekid FROM "%s"`
 	listKeysStmtFmt = `SELECT key FROM "%s"`
 	createTableFmt  = `CREATE TABLE IF NOT EXISTS "%s" (
 		key TEXT UNIQUE NOT NULL PRIMARY KEY,
 		object BLOB,
-		objectnonce TEXT,
-		dek TEXT,
-		deknonce TEXT
+		objectnonce BLOB,
+		dekid BLOB
```
moio, line 29: What pros and cons do you see in using a UUID for the DEK id rather than an incremental counter?

How about storing DEKs in a slice, appending on rotation, and using the slice index as the key identifier instead of the UUID?

pkg/cache/sql/db/client.go
```diff
@@ -72,14 +74,14 @@ type TXClient interface {
 // It should use a key-encryption-key (KEK) to then encrypt the DEK. The encrypted data, nonce used to encrypt the data,
 // encrypted dek, nonce used to decrypt the dek should be returned, in that order. On failure an error should be returned.
```
moio, line 75:
```suggestion
// Encryptor encrypts data with a key which is rotated to avoid wear-out.
```

pkg/cache/sql/db/client.go
```diff
@@ -72,14 +74,14 @@ type TXClient interface {
...
 type Encryptor interface {
-	Encrypt([]byte) ([]byte, []byte, []byte, []byte, error)
+	Encrypt([]byte) ([]byte, []byte, uuid.UUID, error)
```
moio, line 77:
```suggestion
	// Encrypt encrypts the specified data, returning: the encrypted data, the nonce used to encrypt the data, and an ID identifying the key that was used (as it rotates). On failure error is returned instead.
	Encrypt([]byte) ([]byte, []byte, uuid.UUID, error)
```

pkg/cache/sql/db/client.go
```diff
@@ -72,14 +74,14 @@ type TXClient interface {
...
 }
 
 // Decryptor decrypts encrypted data given: some encrypted data, nonce used to encrypt the data, an encrypted
 // data-encryption-key (DEK), nonce used to encrypt the DEK. The KEK is not provided. Decryptor is expected to use its
 // receiver to decrypt the DEK using its KEK.
```
moio, line 82:
```suggestion
// Decryptor decrypts data previously encrypted by Encryptor.
```

pkg/cache/sql/db/client.go
```diff
@@ -72,14 +74,14 @@ type TXClient interface {
...
 type Decryptor interface {
-	Decrypt([]byte, []byte, []byte, []byte) ([]byte, error)
+	Decrypt([]byte, []byte, uuid.UUID) ([]byte, error)
```
moio, line 84:
```suggestion
        // Decrypt accepts a chunk of encrypted data, the nonce used to encrypt it and the ID of the used key (as it rotates). It returns the decrypted data or an error.
	Decrypt([]byte, []byte, uuid.UUID) ([]byte, error)
```

pkg/cache/sql/encryption/encrypt.go
```diff
@@ -1,10 +1,5 @@
-/*
-package encryption provides a struct, Manager, that generates a key-encryption-key, generates and rotates a
-data-encryption-key, and provides functions for encryption and decryption. The key-encryption-key is used so the client
-can store an encrypted version of their data key. This key-hierarchy enables the rotation of data-encryption-keys which
-is intended to protect from cryptanalysis and mitigate impact of leaking a key.
-*/
-
+// package encryption provides encryption and decryption functions, while
+// abstracting away key management concerns.
```
moio, line 2:
```suggestion
// abstracting away key management concerns.
// Current implementation uses AES-GCM encryption, with key rotation, keeping keys in memory
```

pkg/cache/sql/encryption/encrypt.go
```diff
@@ -14,106 +9,99 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/google/uuid"
 	"github.com/pkg/errors"
 )
 
+var (
+	ErrKeyNotFound = errors.New("data key not found")
+	ErrInvalidKey  = errors.New("invalid key")
```
moio, line 18:
```suggestion
```

Seems unused

pkg/cache/sql/encryption/encrypt.go
```diff
@@ -133,50 +121,56 @@ func createGCMCypher(key []byte) (cipher.AEAD, error) {
...
 
-func genRandKey() ([]byte, error) {
-	return genRandByteSlice(keySize)
+	return dek, keyID, nil
 }
 
-func genRandByteSlice(size int) ([]byte, error) {
-	key := make([]byte, size)
-	_, err := rand.Read(key)
+func newRand(d []byte) error {
```
moio, line 169: Optional Nitpick: consider inlining

moio (commented):
Great work @pjbgf , thank you very much for it!

Simpler code, better perf, better test coverage, better security... is there anything more I can really ask for? 😉  really appreciated!

There are a few nitpicks/cosmetic suggestions and only one nontrivial question on pkg/cache/sql/store/store.go - although that's definitely not a deal breaker in any case.

Once that is answered this is ready for my approval and will go to the Frameworks team for another round of review.

pkg/cache/sql/encryption/encrypt.go
pjbgf: The switch was mostly due to the unnecessary performance impact of deferring. But happy to switch back if maintainability is more important in this case.

pkg/cache/sql/store/store.go
pjbgf: Funnily enough, that was my initial approach: the numerical counter part. I pivoted from that mostly to get a step closer to parallel key creation without needing to change the DB schema. I will revert back to that approach instead.

macedogm (commented):
Indeed excellent work, @pjbgf!

Apparently some small updates are still being worked out, so I'll give a general LGTM and avoid approving, as @moio has the fine say.

pjbgf:
I believe I worked through all the feedback on the new commit. The PR description is now updated with uint32 keyID and the latest benchstat. PTAL

moio (approved):
All good, green light from my side!

_(I took the liberty to directly add a few mostly cosmetic commits on top)_

Will merge as soon as we get a 👍 from the frameworks team

Thank you so much for the collaboration

pkg/cache/sql/encryption/encrypt.go
```diff
@@ -17,103 +15,85 @@ import (
...
+// Manager uses AES-GCM encryption and keeps in memory the data encryption
+// keys. The active encryption key is automatically rotated once it has been
+// used over a certain amount of times - defined by maxWriteCount.
 type Manager struct {
-	keyEncryptionKey  []byte      // main key
-	dataEncryptionKey []byte      // current key
-	keyGCMCipher      cipher.AEAD // cipher in gcm mode created from kek
-	writesCounter     int
-	lock              sync.Mutex
+	dataKeys         [][]byte
```
tomleb, line 31: This list only grows and is never shrinked. Is this expected? Won't we run into a memory leak issue if Rancher runs long enough?

tomleb (commented):
Overall looks good.

One concern on memory leaks and one missing lock call.

I'm not super confident with the mocks tests (not your fault) since in the past we've had passing unit tests but the code would fail to run because the integration of sql queries & encryptor was missing some parameters in the SQL. (We've already identified we'd like to have more integration tests for this). So, was this change compiled + tested by running Rancher? That'd make me more confident about the changes here.

pkg/cache/sql/encryption/encrypt.go
moio: Exactly, I am not concerned for that reason.

pkg/cache/sql/encryption/encrypt.go
```diff
@@ -17,103 +15,85 @@ import (
...
-	}
-	dek, err := keyGCMCipher.Open(nil, deknonce, edek, nil)
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to decrypt DEK")
+// Decrypt accepts a chunk of encrypted data, the nonce used to encrypt it and the ID of the used key (as it rotates). It returns the decrypted data or an error.
+func (m *Manager) Decrypt(edata, nonce []byte, keyID uint32) ([]byte, error) {
+	if len(m.dataKeys) <= int(keyID) {
+		return nil, fmt.Errorf("%w: %v", ErrKeyNotFound, keyID)
 	}
+	dek := m.dataKeys[keyID]
```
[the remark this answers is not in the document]
moio: 👍 Nice catch!

pkg/cache/sql/encryption/encrypt.go
tomleb: Ok. Not really familiar with the security aspect of the rotation trigger and key wear-out in general so I'm not able to determine if 150,000 is a good number vs some higher number. My understanding is:
- Higher number means less amount of rotation so less likely to have a memory leak, and I guess instinctively it should be less secure (?)
- Lower number would be the reverse of the above.

> but all those keys need to be accessible for the decryption.

Is it not possible to re-encrypt the data every once in a while? Slowly getting rid of old keys. That way we can keep X number of keys, but X stays finite.

tomleb: And to be clear I understand that this code is under an experimental feature flag, so it's not like we'll see memory leaks for every users. Increasing the key count for rotation seems like the quickest way to make sure it will be impractical to get a noticeable leak. (It still makes me nervous though)

moio: I do not think there are any reasons for nervousness.

Please look at this quantitatively: a key is 32 bytes, therefore to exhaust 1 GB (which is still nowhere near a crazy amount) it'd take 30 _million_ rotations (rounding down for slice inefficiencies).

Even with a very high write rate (50,000 writes/s, which is [at the high end for etcd standalone on good hardware](https://etcd.io/docs/v3.4/op-guide/performance/#benchmarks)) we could have a rotation every 3 seconds - and that might even be untenable with kubeapiserver and the rest in the mix.

Even assuming a rotation every 3 seconds that would take some _90 million_ seconds to use up 1 GB. That's about _three years_ of uptime of a single pod without any kernel update or hardware change or failure of any kind. We are definitely going to run up to several other problems well before that.

On top of that, if the rotation is bumped from 150k to 2^32 (approximately 4B) those numbers will improve by a factor of _28 thousand_!

Even if my math is a bit off I think we can sleep well 😄

pjbgf: Agree with the above - no major concerns here.

I increased the key wear-out var to align with NIST's recommendation, which means we will have a lot less keys to manage (and keep in memory) even in a high throughput deployments.

pkg/cache/sql/encryption/encrypt.go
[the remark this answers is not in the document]
pjbgf: +1 on the nice catch. I made some amendments, removing the counter from the mutex and making all the `dataKeys` ops to be under a `RWMutex`. PTAL

moio (changes requested):

pkg/cache/sql/encryption/encrypt.go
```diff
@@ -118,21 +114,12 @@ func createGCMCypher(key []byte) (cipher.AEAD, error) {
 // the activeKeyCounter exceeds maxWriteCount, the active data key is
 // rotated - before being returned.
 func (m *Manager) fetchActiveDataKey() ([]byte, uint32, error) {
-	m.lock.Lock()
-	defer m.lock.Unlock()
-	m.activeKeyCounter++
-
-	if m.activeKeyCounter >= maxWriteCount {
+	counter := atomic.AddInt64(&m.activeKeyCounter, 1)
```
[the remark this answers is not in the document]
pjbgf: Removed the `atomic` and replaced it with its own lock, given the asymmetrical usage between dataKeys and the counter. Also added an assertion to confirm the counter is being reset correctly.

moio (approved):

moio:
Thanks for the persistence @pjbgf - latest fixes address all my concerns now

FWIW I have also run the integration tests and they are all green https://github.com/rancher/rancher/pull/46015

pkg/cache/sql/encryption/encrypt.go
tomleb: Awesome, I'm good with this, thanks for the numbers, I will be able to sleep well too

tomleb (approved):

PR #81, opened 2024-07-02
sql: when a query fails, add the query text in the error

moio:
This makes debugging easier, especially with integration tests.

Best reviewed commit-by-commit.

*Note* this is based on top of https://github.com/rancher/lasso/pull/82. Ignore the first commit while reviewing

### DIFF
--- a/pkg/cache/sql/informer/indexer.go
+++ b/pkg/cache/sql/informer/indexer.go
@@ -69,7 +69,7 @@ type Store interface {
 
 type DBClient interface {
 	Begin() (db.TXClient, error)
-	QueryForRows(ctx context.Context, stmt transaction.Stmt, params ...any) (*sql.Rows, error)
+	QueryForRows(ctx context.Context, query string, stmt transaction.Stmt, params ...any) (*sql.Rows, error)
 	ReadObjects(rows db.Rows, typ reflect.Type, shouldDecrypt bool) ([]any, error)
 	ReadStrings(rows db.Rows) ([]string, error)
 	Prepare(stmt string) *sql.Stmt
@@ -182,7 +182,7 @@ func (i *Indexer) Index(indexName string, obj any) ([]any, error) {
 		params = append(params, value)
 	}
 
-	rows, err := i.QueryForRows(context.TODO(), stmt, params...)
+	rows, err := i.QueryForRows(context.TODO(), query, stmt, params...)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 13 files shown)


moio:
CI expected to turn green as soon as https://github.com/rancher/lasso/pull/82 is merged

pkg/cache/sql/informer/indexer.go
```diff
@@ -192,7 +192,7 @@ func (i *Indexer) Index(indexName string, obj any) ([]any, error) {
 // ByIndex returns the stored objects whose set of indexed values
 // for the named index includes the given indexed value
 func (i *Indexer) ByIndex(indexName, indexedValue string) ([]any, error) {
-	rows, err := i.QueryForRows(context.TODO(), i.listByIndexStmt, indexName, indexedValue)
+	rows, err := i.QueryForRows(context.TODO(), listByIndexFmt, i.listByIndexStmt, indexName, indexedValue)
```
ericpromislow, line 195: This is the only part I don't get. The call to `i.QueryForRows` adds a second argument that's a template (or format), not the final query.  When I follow it into the implementation function `QueryForRows` at https://github.com/moio/lasso/blob/2bb0987964ad696e592689eef25ee4dba7b853f6/pkg/cache/sql/db/client.go#L115 it looks like the raw template is inserted into the `Error while running query` string, not the final query with the arguments injected into the template.

Am I missing something?

ericpromislow (commented):
Everything else looks fine, but I'm puzzled about how format strings are converted to raw query strings when called from the various functions in `pkg/cache/sql/informer/indexer.go`

pkg/cache/sql/informer/indexer.go
```diff
@@ -69,7 +69,7 @@ type Store interface {
 
 type DBClient interface {
 	Begin() (db.TXClient, error)
-	QueryForRows(ctx context.Context, stmt transaction.Stmt, params ...any) (*sql.Rows, error)
+	QueryForRows(ctx context.Context, query string, stmt transaction.Stmt, params ...any) (*sql.Rows, error)
```
tomleb, line 72: nit: Since we have both query and stmt, could we document how they differ?

tomleb (approved):
LGTM.

I assume the intention here is to get _some_ information from the failing query rather than _all_ information though I'd like a confirmation on Eric's comment.

moio:
Gentlemen, thanks for the feedback.

While reasoning about how to address it I came up with a different approach which is more thorough, more precise (@ericpromislow 's point), less invasive and (hopefully) easier to read (@tomleb 's point).

As the change is completely different I am closing this in favor of https://github.com/rancher/lasso/pull/83

PR #82, opened 2024-07-02
sql: listoption_indexer: handle the case with one empty filter correctly

moio:
Integration tests revealed this case result in error 500 because an incorrect query is produced in this case (`WHERE ()`).

This adds a test and addresses the case.


tomleb (approved):
LGTM

ericpromislow (approved):
New test fails without implementation change, passes with. LGTM

PR #83, opened 2024-07-05
sql: when a query fails, add the query text in the error (take 2)

moio:
This makes debugging easier, especially with integration tests.

Compared to the [earlier attempt](https://github.com/rancher/lasso/pull/81) this is more thorough, more precise, less invasive and (hopefully) easier to read.

### DIFF
--- a/pkg/cache/sql/informer/informer_test.go
+++ b/pkg/cache/sql/informer/informer_test.go
@@ -9,8 +9,12 @@ import (
 	"github.com/golang/mock/gomock"
 	"github.com/rancher/lasso/pkg/cache/sql/partition"
 	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/tools/cache"
 )
 
 //go:generate mockgen --build_flags=--mod=mod -package informer -destination ./informer_mocks_test.go github.com/rancher/lasso/pkg/cache/sql/informer ByOptionsLister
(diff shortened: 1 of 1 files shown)


tomleb (approved):
LGTM.

A nit would be the increased usage of `errors.Wrap*` instead of the built-in `fmt.Errorf` to wrap errors. Currently I only see `errors.Wrap*` being used in `pkg/cache/sql`, it seems like `fmt.Errorf` was already favored in the rest of lasso. Not sure how strongly we feel about this though.

moio:
Happy to change them all to `fmt.Errorf` + `%w` if that helps

moio:
@tomleb done

pkg/cache/sql/store/store.go
```diff
@@ -203,7 +216,7 @@ func (s *Store) List() []any {
 func (s *Store) ListKeys() []string {
 	rows, err := s.QueryForRows(context.TODO(), s.listKeysStmt)
 	if err != nil {
-		fmt.Printf("Unexpected error in store.ListKeys: %v\n", err)
+		fmt.Printf("Unexpected error in store.ListKeys: %v\n", fmt.Errorf("while executing query: %s got error: %w", s.listKeysQuery, err))
```
bigkevmcd, line 219: Why not compose this as a single string?

It's now using `%v` which will trigger `.Error()` on the `err` so it could just be a single `fmt.Printf(...)`?

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -246,15 +263,15 @@ func (l *ListOptionIndexer) ListByOptions(ctx context.Context, lo ListOptions, p
...
-			stmt += fmt.Sprintf("(%s)", clause)
+			query += fmt.Sprintf("(%s)", clause)
 			if index == len(whereClauses)-1 {
 				break
 			}
-			stmt += " AND\n    "
+			query += " AND\n    "
 		}
-		// stmt += strings.Join(whereClauses, " AND\n    ")
+		// query += strings.Join(whereClauses, " AND\n    ")
```
bigkevmcd, line 274: Do we need this line to remain commented out?

pkg/cache/sql/store/store.go
```diff
@@ -107,7 +120,7 @@ func (s *Store) upsert(key string, obj any) error {
 
 	err = s.Upsert(tx, s.upsertStmt, key, obj, s.shouldEncrypt)
 	if err != nil {
-		return err
+		return fmt.Errorf("while executing query: %s got error: %w", s.upsertQuery, err)
```
bigkevmcd, line 123: This looks like it could be an error type, similar to https://pkg.go.dev/io/fs#PathError ?

```go
type QueryError struct {
  QueryString string
  Err error
}

func (e *QueryError) Error() string { return "while executing query: " + e.QueryString + " got error: " + e.Err.Error() }

func (e *QueryError) Unwrap() error { return e.Err }
```

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -144,7 +152,7 @@ func (l *ListOptionIndexer) afterUpsert(key string, obj any, tx db.TXClient) err
 			logrus.Errorf("cannot index object of type [%s] with key [%s] for indexer [%s]: %v", l.GetType().String(), key, l.GetName(), err)
 			cErr := tx.Cancel()
 			if cErr != nil {
-				return errors.Wrap(err, cErr.Error())
+				return fmt.Errorf("could not cancel transaction: %s while recovering from error: %w", cErr.Error(), err)
 			}
```
bigkevmcd, line 156: The use of `%s` will trigger the call to `cErr.Error()` so you don't need to do it explicitly.

> 4. If an operand implements the error interface, the Error method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any).

pkg/cache/sql/informer/listoption_indexer.go
moio: Addressed in current master

pkg/cache/sql/store/store.go
moio: Right, thanks for the observation. Unfortunately commits in this PR were merged together as part of #84, so I am starting [a separate PR](https://github.com/rancher/lasso/pull/90) to address this feedback

pkg/cache/sql/store/store.go
moio: Right, thanks for the observation. Unfortunately commits in this PR were merged together as part of #84, so I am starting [a separate PR](https://github.com/rancher/lasso/pull/90) to address this feedback

pkg/cache/sql/informer/listoption_indexer.go
moio: Right, thanks for the observation. Unfortunately commits in this PR were merged together as part of #84, so I am starting [a separate PR](https://github.com/rancher/lasso/pull/90) to address this feedback

PR #84, opened 2024-07-05
sql: teach ListByOptions to return the total object count

moio:
*Problem*: ListByOptions is currently unable to return the total object count (it will only return the requested page), and Steve needs it for the UI.

This PR adds the lasso bits, the accompanying Steve PR https://github.com/rancher/steve/pull/236 adds the Steve specific bits.

FWIW I have also run the integration tests and they are all green https://github.com/rancher/rancher/pull/46015

**Best reviewed commit-by-commit.**

**Please note this contains commits from https://github.com/rancher/lasso/pull/83, disregard the first two commits.**

**Please note this PR should be merged before https://github.com/rancher/steve/pull/236.**

### DIFF
--- a/pkg/cache/sql/db/client_test.go
+++ b/pkg/cache/sql/db/client_test.go
@@ -296,6 +296,96 @@ func TestQueryStrings(t *testing.T) {
 	}
 }
 
+func TestReadInt(t *testing.T) {
+	type testCase struct {
+		description string
+		test        func(t *testing.T)
+	}
+
+	var tests []testCase
+
+	testResult := 42
+	tests = append(tests, testCase{description: "One row, no errors", test: func(t *testing.T) {
+		c := SetupMockConnection(t)
+		e := SetupMockEncryptor(t)
+		d := SetupMockDecryptor(t)
+		r := SetupMockRows(t)
+		r.EXPECT().Next().Return(true)
+		r.EXPECT().Scan(gomock.Any()).Do(func(a ...any) {
+			p := a[0].(*int)
+			*p = testResult
+		})
+		r.EXPECT().Err().Return(nil)
+		r.EXPECT().Close().Return(nil)
+		client := SetupClient(t, c, e, d)
+		result, err := client.ReadInt(r)
+		assert.Nil(t, err)
+		assert.Equal(t, 42, result)
+	},
+	})
+	tests = append(tests, testCase{description: "One row, Scan error", test: func(t *testing.T) {
+		c := SetupMockConnection(t)
+		e := SetupMockEncryptor(t)
+		d := SetupMockDecryptor(t)
+		r := SetupMockRows(t)
+		r.EXPECT().Next().Return(true)
+		r.EXPECT().Scan(gomock.Any()).Return(fmt.Errorf("error"))
+		r.EXPECT().Close().Return(nil)
+		client := SetupClient(t, c, e, d)
+		_, err := client.ReadInt(r)
+		assert.NotNil(t, err)
+	},
+	})
+	tests = append(tests, testCase{description: "One row, Err() error", test: func(t *testing.T) {
+		c := SetupMockConnection(t)
+		e := SetupMockEncryptor(t)
+		d := SetupMockDecryptor(t)
+		r := SetupMockRows(t)
+		r.EXPECT().Next().Return(true)
+		r.EXPECT().Scan(gomock.Any()).Do(func(a ...any) {
+			a[0] = testResult
+		})
+		r.EXPECT().Err().Return(fmt.Errorf("error"))
+		r.EXPECT().Close().Return(nil)
+		client := SetupClient(t, c, e, d)
+		_, err := client.ReadInt(r)
+		assert.NotNil(t, err)
+	},
+	})
+	tests = append(tests, testCase{description: "One row, Close() error", test: func(t *testing.T) {
+		c := SetupMockConnection(t)
+		e := SetupMockEncryptor(t)
+		d := SetupMockDecryptor(t)
+		r := SetupMockRows(t)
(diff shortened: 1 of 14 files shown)


pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -285,76 +305,94 @@ func (l *ListOptionIndexer) ListByOptions(ctx context.Context, lo ListOptions, p
...
+		rows, err = l.QueryForRows(ctx, countStmt, countParams...)
+		if err != nil {
+			return nil, 0, "", errors.Wrapf(err, "Error while executing query:\n %s", countQuery)
+		}
+		total, err = l.ReadInt(rows)
+		if err != nil {
+			return nil, 0, "", err
+		}
+	} else {
+		total = len(items)
```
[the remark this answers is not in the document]
moio: addressed in 27e58f6

tomleb (approved):
All in all looks good to me.

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -285,76 +305,92 @@ func (l *ListOptionIndexer) ListByOptions(ctx context.Context, lo ListOptions, p
...
-		return nil, "", err
+		return nil, 0, "", err
+	}
+
+	total := len(items)
+	// if limit or offset were set, execute counting of all rows
+	if limit > 0 || offset > 0 {
+		countStmt := l.Prepare(countQuery)
+		defer l.CloseStmt(countStmt)
+		rows, err = l.QueryForRows(ctx, countStmt, countParams...)
```
tomleb, line 378: Actually, let me check but I think both those query for rows need to be in a transaction and its not clear to me that they are.

pkg/cache/sql/db/client_test.go
```diff
@@ -296,6 +296,96 @@ func TestQueryStrings(t *testing.T) {
...
+	type testCase struct {
+		description string
+		test        func(t *testing.T)
+	}
+
+	var tests []testCase
+
+	testResult := 42
+	tests = append(tests, testCase{description: "One row, no errors", test: func(t *testing.T) {
+		c := SetupMockConnection(t)
```
ericpromislow, line 309: Lots of repetition here ... lines 309 through 312 and 320 can be done in one helper func that returns `(*MockRows, *Client)` all through this test. And then the tests are more readable because the setup is minimized.

MbolotSuse (changes requested):
Most changes look good. Big concerns here are:
- As @tomleb pointed out, a transaction might help us to produce more consistency between the "results" query and the "count" query
-  I'm a bit unsure about returning the whole query in our error statements. I'd prefer if we rendered a more direct error statement so that we have more control in what is potentially being returned here.
- I'm a bit concerned about a possible SQL injection with the use of the () on the new count statement, but I definitely could be wrong there.

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -285,76 +305,92 @@ func (l *ListOptionIndexer) ListByOptions(ctx context.Context, lo ListOptions, p
...
+		if err != nil {
+			return nil, 0, "", errors.Wrapf(err, "Error while executing query:\n %s", countQuery)
```
ericpromislow, line 380: Thought we prefer `fmt.Errorf(...%w...)` to `errors.Wrapf` ?

ericpromislow (commented):
Just nits that shouldn't hold up merging this.

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -285,76 +305,92 @@ func (l *ListOptionIndexer) ListByOptions(ctx context.Context, lo ListOptions, p
...
+			query += "\n  ORDER BY f.\"metadata.name\" ASC "
 		}
 	}
 
-	// 3- Pagination: LIMIT clause (from lo.Pagination and/or lo.ChunkSize/lo.Resume)
+	// 4- Pagination: LIMIT clause (from lo.Pagination and/or lo.ChunkSize/lo.Resume)
+
+	// before proceeding, save a copy of the query and params without LIMIT/OFFSET
+	// for COUNTing all results later
+	countQuery := fmt.Sprintf("SELECT COUNT(*) FROM (%s)", query)
```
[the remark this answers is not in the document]
ericpromislow: I saw this as well, and it looked like `query` is safe, but I'd prefer using a safer statement-prep form as a matter of style

pkg/cache/sql/informer/listoption_indexer.go
tomleb: I'd still like some input from @moio about this. While we're not writing to the DB, it seems like we should be returning a consistent result. However, it might not be critical if the worse that can happen is the UI receiving a more up-to-date value.. I'm okay with merging this as is to get things moving.

tomleb (approved):

MbolotSuse (approved):
Approving in the interest of unblocking other work, though we will need to adjust some of this in the future.

ericpromislow (approved):
ok

pkg/cache/sql/informer/listoption_indexer.go
moio: I will fix that in a follow-up PR, thanks for spotting

pkg/cache/sql/informer/listoption_indexer.go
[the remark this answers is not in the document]
moio: @MbolotSuse not from my understanding, as query is (supposed to be) independently safe.

@ericpromislow I do not understand your proposal. `query` needs to be string-built, as opposed to prepared, because table and column names are variable - and so does `countQuery` by consequence.

How would you propose to use a prepared statement in this case?

PR #86, opened 2024-07-05
Add test that will break if UnsafeSet breaks

moio:
SQLite based caching uses an `Informer` that unsafely sets the `Indexer` as the ability to set it is not present in client-go at the moment.

Long term, we look forward contribute a patch to client-go to make that configurable.

Until then, we are adding this canary test that will panic in case the indexer cannot be set.


bigkevmcd (commented):
Can I suggest adding the comment in the PR to the `TestUnsafeSet` so that other folks who read it understand _why_ ?

moio:
Good point, added, thanks @bigkevmcd

bigkevmcd (approved):

MbolotSuse (approved):

PR #91, opened 2024-07-31
Fix datarace when reading handlers in SharedHandler

alexandreLamarre:


MbolotSuse (changes requested):
This seems to be close to what the code was previously doing: https://github.com/rancher/lasso/commit/a3ff81636758af01724b91631354169b460e8259. I think we need more consideration on this one before making changes.

ericpromislow:
Before the code was holding the lock while it ran each handler (and there are various metrics in the loop that suggest people have been concerned with how long it takes to update an object by applying an array of (mutating webhook?) handlers to it. Now the use of the lock suggests that it's waiting for the array of handlers to be built, and then it uses it.

But the fact of the race condition in the helm-lock tests suggest that something else is modifying the array while the lasso *sharedhandler* loops through it, so it's safer to make a shallow copy of it and loop through that. If the underlying
array changes while the `OnChange` method loops through, the only loss is that our mutation might be out-dated.

ericpromislow (approved):
Well, I wrote this code, but I think it's correct, and not onerously expensive.

tomleb:
We could avoid making a copy for every `OnChange` calls by creating a new handler slice when a new handler is added.

ericpromislow:
This needs to be redone against `main`, and then I'll request two other reviewers who work here and aren't me

ericpromislow:
Does this really need 19 commits changing 49 files? And you also want to merge this against `main`, not `master` (prob that's why so many commits).

PR #93, opened 2024-08-14
sql: handle listable unwatchable resources gracefully

moio:
Fixes https://github.com/rancher/rancher/issues/46030

Goes together with https://github.com/rancher/steve/pull/263

### DIFF
--- a/pkg/cache/sql/informer/factory/informer_factory.go
+++ b/pkg/cache/sql/informer/factory/informer_factory.go
@@ -30,19 +30,15 @@ type CacheFactory struct {
 
 	newInformer newInformer
 
-	cache map[schema.GroupVersionKind]*informer.Informer
+	cache map[schema.GroupVersionKind]informer.Informer
 }
 
-type newInformer func(client dynamic.ResourceInterface, fields [][]string, gvk schema.GroupVersionKind, db sqlStore.DBClient, shouldEncrypt bool, namespace bool) (*informer.Informer, error)
+type newInformer func(client dynamic.ResourceInterface, fields [][]string, gvk schema.GroupVersionKind, db sqlStore.DBClient, shouldEncrypt bool, namespace bool) (informer.Informer, error)
 
 type DBClient interface {
 	informer.DBClient
 	sqlStore.DBClient
 	connector
-}
-
-type Cache struct {
-	informer.ByOptionsLister
 }
 
 type connector interface {
@@ -81,7 +77,7 @@ func NewCacheFactory() (*CacheFactory, error) {
 	return &CacheFactory{
 		wg:          wait.Group{},
 		stopCh:      make(chan struct{}),
-		cache:       map[schema.GroupVersionKind]*informer.Informer{},
+		cache:       map[schema.GroupVersionKind]informer.Informer{},
 		encryptAll:  os.Getenv(EncryptAllEnvVar) == "true",
 		dbClient:    dbClient,
 		newInformer: informer.NewInformer,
@@ -89,10 +85,10 @@ func NewCacheFactory() (*CacheFactory, error) {
 }
 
 // CacheFor returns an informer for given GVK, using sql store indexed with fields, using the specified client
-func (f *CacheFactory) CacheFor(fields [][]string, client dynamic.ResourceInterface, gvk schema.GroupVersionKind, namespaced bool) (Cache, error) {
+func (f *CacheFactory) CacheFor(fields [][]string, client dynamic.ResourceInterface, gvk schema.GroupVersionKind, namespaced bool) (informer.Informer, error) {
 	result, ok := f.getCacheIfExists(gvk)
 	if ok {
-		return Cache{ByOptionsLister: result}, nil
+		return result, nil
 	}
 
 	f.informerCreateLock.Lock()
@@ -102,22 +98,22 @@ func (f *CacheFactory) CacheFor(fields [][]string, client dynamic.ResourceInterf
 	shouldEncrypt := f.encryptAll || encryptResourceAlways
 	i, err := f.newInformer(client, fields, gvk, f.dbClient, shouldEncrypt, namespaced)
 	if err != nil {
-		return Cache{}, err
+		return nil, err
 	}
 
 	if f.cache == nil {
-		f.cache = make(map[schema.GroupVersionKind]*informer.Informer)
+		f.cache = make(map[schema.GroupVersionKind]informer.Informer)
 	}
 	f.cache[gvk] = i
 	f.wg.StartWithChannel(f.stopCh, i.Run)
 	if !cache.WaitForCacheSync(f.stopCh, i.HasSynced) {
-		return Cache{}, fmt.Errorf("failed to sync SQLite Informer cache for GVK %v", gvk)
+		return nil, fmt.Errorf("failed to sync SQLite Informer cache for GVK %v", gvk)
 	}
 
-	return Cache{ByOptionsLister: i}, nil
+	return i, nil
 }
 
-func (f *CacheFactory) getCacheIfExists(gvk schema.GroupVersionKind) (*informer.Informer, bool) {
+func (f *CacheFactory) getCacheIfExists(gvk schema.GroupVersionKind) (informer.Informer, bool) {
 	f.informerCreateLock.RLock()
 	defer f.informerCreateLock.RUnlock()
 
@@ -141,7 +137,7 @@ func (f *CacheFactory) Reset() error {
 
 	close(f.stopCh)
 	f.stopCh = make(chan struct{})
-	f.cache = make(map[schema.GroupVersionKind]*informer.Informer)
+	f.cache = make(map[schema.GroupVersionKind]informer.Informer)
 	err := f.dbClient.NewConnection()
 	if err != nil {
 		return err
--- a/pkg/cache/sql/informer/factory/informer_factory_test.go
+++ b/pkg/cache/sql/informer/factory/informer_factory_test.go
@@ -64,14 +64,10 @@ func TestInformerFor(t *testing.T) {
 		sii := NewMockSharedIndexInformer(gomock.NewController(t))
 		sii.EXPECT().HasSynced().Return(true)
 		sii.EXPECT().Run(gomock.Any()).MinTimes(1)
-		i := &informer.Informer{
-			// need to set this so Run function is not nil
-			SharedIndexInformer: sii,
-		}
-		expectedC := Cache{
-			ByOptionsLister: i,
-		}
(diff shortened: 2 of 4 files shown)


pkg/cache/sql/informer/factory/informer_factory.go
```diff
@@ -149,3 +164,23 @@ func (f *CacheFactory) Reset() error {
...
+	}
+
+	for _, resource := range apiResourceList.APIResources {
+		if resource.Kind == gvk.Kind {
+			for _, verb := range resource.Verbs {
+				if verb == "watch" {
+					return nil
+				}
+			}
+			return &UnwatchableError{gvk}
```
bigkevmcd, line 182: You could use `slices.Contains(resource.Verbs, "watch")` ?

And I don't think you need to return `&WatchableError{gvk}` here, because if none of the `apiResourceList.APIResources` have a "watch" verb, then it will just drop out to the return at the end of the function?

MbolotSuse: Agree with the points above. I think that we should still break at this point to avoid iterating over the other GVKs once we have located the target one.

pkg/cache/sql/informer/factory/informer_factory.go
```diff
@@ -149,3 +164,23 @@ func (f *CacheFactory) Reset() error {
 
 	return nil
 }
+
+// checkWatchability checks whether the specified gvk can be watched by this cache factory
+func (f *CacheFactory) checkWatchability(gvk schema.GroupVersionKind) error {
```
MbolotSuse, line 169: The performance of this could be a bit rough because this will list all of the resources in a group. In addition, it doesn't cache the response for the given resource type, so if the UI (for example) attempts to watch this resource it will continually attempt to evaluate each of these types. 

I'm wondering if alternatively we could use the `SetWatchErrorHandler` method of the [SharedInformer](https://pkg.go.dev/k8s.io/client-go/tools/cache#SharedInformer) and to check for a [IsMethodNotSupported](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#IsMethodNotSupported) error. This might be a little more complicated, but I think this also could be a bit more accurate since it's parsed from the direct response of the APIserver rather than the openapi schema (which may be inaccurate). WDYT?

MbolotSuse (commented):
One suggestion in addition to the current comments.

pkg/cache/sql/informer/factory/informer_factory.go
bigkevmcd: While I agree that caching would be useful to avoid the API hit, there shouldn't be many `APIResources` in the `apiResourceList`, this is basically getting all the CRDs in the group?

bigkevmcd: If we're going to assume that the OpenAPI specs maybe wrong, we should figure out a plan to fix this.

bigkevmcd:
There's a test missing for the case where `watch` is not allowed?

pkg/cache/sql/informer/factory/informer_factory.go
tomleb: >While I agree that caching would be useful to avoid the API hit, there shouldn't be many APIResources in the apiResourceList, this is basically getting all the CRDs in the group?

That's correct. For a default rancher install the group/version pair with the most resources is `management.cattle.io/v3` with ~50 entries. Don't know how many more resources you could get (eg: with crossplane). Though that will be multiplied by how many cache factory UI is asking (IIUC).

The client-go library has code for caching those calls if that's something we deem important enough: (eg: https://github.com/kubernetes/client-go/blob/master/discovery/discovery_client.go#L105 or https://github.com/kubernetes/client-go/tree/master/discovery/cached). Would have to make sure we invalidate the cache once in a while because the set of api resources can change over time (eg: new CRD, new apiservice, etc).

> rather than the openapi schema (which may be inaccurate)

What do you mean by openapi here? OpenAPI isn't being hit by the code below, just the discovery API.

As far as accuracy goes, it is possible to implement a custom apiserver, that lists `watch` as a supported verb, aggregate it to the main apiserver and have a watch fail. But at this point the contract between main apiserver / custom apiserver is broken so I doubt we could even trust the error response (the buggy apiserver could just send a 500 instead of methodnotsupported :shrug: .

MbolotSuse: > While I agree that caching would be useful to avoid the API hit, there shouldn't be many APIResources in the apiResourceList, this is basically getting all the CRDs in the group?

I think @tomleb  gave a good response to you below, but there can be a couple of resources in the same group, and keep in mind that the call will be made for each of them. 

> I had a deeper look, we're already caching the Cache for the gvk?

Only if the cache was succesfully created: https://github.com/rancher/lasso/blob/6ae667e0fe659adc9f5b77c3adac1e2780555465/pkg/cache/sql/informer/factory/informer_factory.go#L103. My comment here is dealing with the failure case which is not cached (at least as far as I can see).

> If we're going to assume that the OpenAPI specs maybe wrong, we should figure out a plan to fix this.

@tomleb explained this part as well, but an API server can be misconfigured to identify methods it doesn't actually serve. If this is done by a third party (or closed source) application we would have no recourse to fixing this. 

Now it could be argued that my solution depends on us being able to figure out if watch is/is not supported, so there's some room to discuss there, but given that the user can't disable this logic I think that's worth considering.

> OpenAPI isn't being hit by the code below, just the discovery API.

My fault, this would just be the discovery endpoints as @tomleb noted.

> I doubt we could even trust the error response (the buggy apiserver could just send a 500 instead of methodnotsupported 🤷 .)

This is a good point, however, we could parse out more than this specific error (i.e. look for retryable error codes and only retry in those cases, so don't retry with a generic 500). This would still require the APIServer to configure itself correctly, but this would limit our retry action (which is a call to an API) to cases where we had some indication that the issue could be resolved.

moio: Valid comments, thank you.

I pushed a completely different approach now so I am resolving this.

moio:
@tomleb I would like your opinion on behalf of the Team about the general approach to take here.

The initial idea behind this PR was to check whether a resource is watchable or not before starting its Informer. That of course incurs the cost of such checking, albeit it is only once per Kind, only when it is displayed.

A different approach was suggested by Michael and is presented in the current version of the PR: start the Informer in all cases, allowing it to fail (and back-off) for unwatchable resources. In so doing avoid to spam logs, and keep track of failures in order to later be able to react on `List` calls appropriately. _(Note I ended up coding all of it almost in full because I was not able to tell all consequences just by thinking through them)_

The core of this new approach is setting an Informer-specific error handler (which silences debug log lines and sets an `unwatchable` flag) and a new event handler (which resets the flag as it implies data is flowing):

[`24dbf99` (#93)](https://github.com/rancher/lasso/pull/93/commits/24dbf99def1f8ab89ff0bcb6018072d4d0f2cfcf#diff-3ec59afaaacb5f46b0c39dd3eeaf7572c8877d95e11a2afcd40d56fd1551ab7fL68-R113)

Arguably, this is a less straightforward solution. It may have a small performance advantage. It will work if resources change from watchable to unwatchable and vice versa, although that may also never happen. It may behave a bit more accurately in some misconfiguration corner cases, but those may also never happen or be supported anyway.

Personally, at the end of the day, I think the maintainers' opinion on what they prefer to support going forward is the most relevant factor here.

What would you choose?

tomleb:
> @tomleb I would like your opinion on behalf of the Team about the general approach to take here.
> 
> The initial idea behind this PR was to check whether a resource is watchable or not before starting its Informer. That of course incurs the cost of such checking, albeit it is only once per Kind, only when it is displayed in the UI.
> 
> A different approach was suggested by Michael and is presented in the current version of the PR: start the Informer in all cases, allowing it to fail (and back-off) for unwatchable resources. In so doing avoid to spam logs, and keep track of failures in order to later be able to react on `List` calls appropriately. _(Note I ended up coding all of it almost in full because I was not able to tell all consequences just by thinking through them)_
> 
> The core of this new approach is setting an Informer-specific error handler (which silences debug log lines and sets an `unwatchable` flag) and a new event handler (which resets the flag as events flowing imply that watch works):
> 
> [`24dbf99` (#93)](https://github.com/rancher/lasso/pull/93/commits/24dbf99def1f8ab89ff0bcb6018072d4d0f2cfcf#diff-3ec59afaaacb5f46b0c39dd3eeaf7572c8877d95e11a2afcd40d56fd1551ab7fL68-R113)
> 
> Arguably, this is a less straightforward solution. It may have a small performance advantage. It will work if resources change from watchable to unwatchable and vice versa, although that may also never happen. It may behave a bit more accurately in some misconfiguration corner cases, but those may also never happen or be supported anyway.
> 
> Personally, at the end of the day, I think the maintainers' opinion on what they prefer to support going forward is the most relevant factor here.
> 
> What would you choose?

@moio After a discussion with Michael, we've concluded we'd like to go with the following approach:
1. Checking whether a GVK can be watched should be done using the `APIResource` resource because that's the standard way of checking this and we're [already doing this in Steve](https://github.com/rancher/steve/blob/5ec7c773df13ecfade2d63e9a82c309b54be33c5/pkg/controllers/schema/schemas.go#L111-L127) by using the verbs attribute that's already cached. We should be able to do this check in Steve instead of Lasso (note: doing it in lasso would be fine too but my opinion is that it's simpler in Steve due to `attributes.Verbs()` being available.
2. We'd like to better handle a resource failing to being watched in lasso. Ideally we'd be able to prevent the watcher to not hammer the kube-apiserver (if that's the behavior, at this point I'm not sure). Some kind of rate limiter there would be nice to prevent this + I suspect it would also prevent some log spam.
Let me know what you think (and apologies for making you change the solution a third time in case you agree with the plan above).

moio:
@tomleb 

1. I am open to give a try to the Steve-based approach - the reasoning makes sense to me at a high level.
2. The retry mechanism is already present, as it is built into client-go, and it already implements back-off/rate-limiting to not overwhelm kube-apiserver:

https://github.com/kubernetes/client-go/blob/46965213e4561ad1b9c585d1c3551a0cc8d3fcd6/tools/cache/reflector.go#L306-L310

The back-off parameters are, FMPOV, pretty reasonable as defaults:

https://github.com/kubernetes/client-go/blob/46965213e4561ad1b9c585d1c3551a0cc8d3fcd6/tools/cache/reflector.go#L233-L236

If ever we wanted to change them, unfortunately, we'd have to patch client-go itself - because they are all hardcoded. And the patch should ideally be upstream - which is a bit of work to make happen.

Therefore, for this case, I'd say it's not worth the hassle once 1. is actually implemented, as that should cover the very vast majority of cases.


Expect a new Steve PR from me, hopefully in the next days. Thanks

PR #94, opened 2024-08-14
Escape underscores and percent signs in filter matches in Vai

maxsokolovsky:
Issue: https://github.com/rancher/rancher/issues/46410

This PR changes the logic of how `ListOptionIndexer` builds a SQL query from filters.

The indexer will now escape two characters: `_` and `%`. As a result, the SQL engine interprets them literally within a `LIKE` statement, not as special characters as it normally does. For example, it treats `_` as a placeholder for a single character. We don't want that. Now it escapes it and expects the literal underscore.

I adjusted existing unit tests but haven't added any new cases, because the current test infrastructure relies on mocked values too much, so I can't assert on returned values from SQL, since those are always mocked.

We will need proper integration tests, which is a separate issue.


bigkevmcd:
We should have integration tests for this code?

maxsokolovsky:
@bigkevmcd, definitely. It's just that this effort may be outside the scope of this issue. @moio, what do you think?

bigkevmcd:
@maxsokolovsky I get that, and maybe in this case, yes, but I don't see much value in the mocked version, it's really just testing the mocks?

tomleb:
I'm not a fan of those mock tests and we've had discussion about them during the first iteration. We've added a bullet point (now [here](https://github.com/rancher/rancher/issues/46348)) about adding integration tests. Also iirc in the very first iteration that @moio worked on, they were reusing tests from upstream's client-go, which I think is a great idea and we should bring that back (also noted in the link).

It seems like even though we have that as a todo, I doubt they'll end up being a priority so maybe we should start adding pieces of non-mock tests as we work through the rest of the tasks. Would it require a lot of effort to add a couple of tests that doesn't just check the query string, but actually runs it with an in-memory sqlite database? @maxsokolovsky

MbolotSuse:
@tomleb/ @bigkevmcd  we have an open issue to backfill tests for lasso: https://github.com/rancher/rancher/issues/45243. If you want to add a comment to that issue requesting that this be part of the backfill, that would be great. However, I'm not in favor of asking @maxsokolovsky to design/implement a way to do this in order for him to submit this small PR.

moio (approved):

moio:
I agree with @MbolotSuse, this is an important topic but should not be tackled in the context of this PR.

@tomleb we have control over importance - there is consensus we deem it important, therefore we will make it happen at the expense of something else.

I approved this PR, and plan to merge it as soon as anybody else adds another approval.

moio:
@tomleb can you please take a look and approve?

maxsokolovsky:
@moio, there is a [PR](https://github.com/rancher/lasso/pull/101) that adds integration tests. If you want, we can wait until that one is merged, and I will add test cases in this PR.

moio:
@maxsokolovsky I like that plan

maxsokolovsky:
Added a couple of cases to our new integration tests.

moio (approved):
LGTM! Thanks for this contribution

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -417,12 +418,15 @@ func (l *ListOptionIndexer) buildORClauseFromFilters(orFilters OrFilter) (string
...
+		orWhereClause += fmt.Sprintf(`f."%s" %s ? ESCAPE '\'`, columnName, opString)
 		format := strictMatchFmt
 		if filter.Partial {
 			format = matchFmt
 		}
-		params = append(params, fmt.Sprintf(format, filter.Match))
+		match := filter.Match
+		match = strings.ReplaceAll(match, `_`, `\_`)
+		match = strings.ReplaceAll(match, `%`, `\%`)
+		params = append(params, fmt.Sprintf(format, match))
```
tomleb, line 429: Should we support filtering fields that contain `\`?

Right now afaik it doesn't:

```
sqlite> create table T(t text);
sqlite> insert into T(t) values ('my\windows\path');
sqlite> select * from T where t LIKE 'my\windows\path' ESCAPE '\';
sqlite> select * from T where t LIKE 'my\\windows\\path' ESCAPE '\';
my\windows\path
```

Maybe we need. Would also be worth it to add a test case for this.

```suggestion
		match := filter.Match
		match = strings.ReplaceAll(match, `\`, `\\`)
		match = strings.ReplaceAll(match, `_`, `\_`)
		match = strings.ReplaceAll(match, `%`, `\%`)
		params = append(params, fmt.Sprintf(format, match))
```

maxsokolovsky: You don't need to use `ESCAPE` in your first query - then sqlite will return that row.
The backslash isn't special in sqlite, unlike the underscore and percent sign.

This PR aims to work around the special behavior of the LIKE operator when used with strings containing `_` and `%`.
Here are some docs describing it:
https://www.sqlite.org/lang_expr.html#the_like_glob_regexp_match_and_extract_operators

tomleb: I'm not sure what you mean by 

> You don't need to use ESCAPE in your first query - then sqlite will return that row.

In the context of this PR. The PR adds `ESCAPE` to the query unconditionally. So my example shows what the query will look like, meaning it will have where clauses with `ESCAPE`. Right now, this doesn't allow matching fields with `\` in them.

I added a quick test in there (I'm only showing the test but I added the rest of the code necessary to make that work):

```go
		{
			name: "match backslash",
			filters: orFiltersForFilters(informer.Filter{
				Field:   []string{`metadata`, `annotations[somekey]`},
				Match:   `my\backslash\path`,
				Op:      informer.Eq,
				Partial: false,
			}),
			wantNames: []string{"backslash"},
		},

```

and this fails with:

```
                        Error:          Not equal:                                                                                                                                                                                                                                                                     
                                        expected: 1                                                                                                                                                                                                                                                                    
                                        actual  : 0                                                                                                                                                                                                                                                                    
                        Test:           TestIntegrationSuite/TestSQLCacheFilters/match_backslash
```

because it can't find the configmap.

tomleb: Adding

```go
match = strings.ReplaceAll(match, `\`, `\\`)
```

makes that test succeed.

maxsokolovsky: Ah, I see, good point. Using ESCAPE indiscriminately in the query ignores the back slash itself when it's needed.
I think we could change that as you suggest.

tomleb (approved):

moio (approved):

moio:
Nice job, thanks for all the collaboration!

PR #95, opened 2024-08-22
feat(metrics): improve lasso_controller_total_cached_object

aruiz14:
Issue: https://github.com/rancher/rancher/issues/46783

Includes the following improvements:
* It's updated every minute with the real value, instead of at the start.
* Can be split by context, by calling `WithContextID` over the `context.Context` passed to `Start`, see https://github.com/rancher/rancher/pull/46784
* Metric is no longer orphaned when the cache is stopped.

Example:
Screenshot 2024-08-22 at 08 47 33
(note for every `ctx` were set from [rancher](https://github.com/rancher/rancher/pull/46784) and can be changed)


moio:
It is my understanding that all current lasso metrics are gated behind the `CATTLE_PROMETHEUS_METRICS` environment variable.

Should `startMetricsCollection` honor it too?

pkg/metrics/metrics.go
```diff
@@ -14,11 +15,25 @@ const (
 	handlerNameLabel    = "handler_name"
 	hasErrorLabel       = "has_error"
 
+	contextLabel = "ctx"
 	groupLabel   = "group"
 	versionLabel = "version"
 	kindLabel    = "kind"
+
+	contextIDKey = "cache_context_id"
```
bigkevmcd, line 23: While it's very unlikely to cause a clash, the best practices  seem to have formed around non-string keys for context keys?

bigkevmcd (changes requested):
I'd like to see tests added for this?

Might wanna checkout https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/testutil

pkg/cache/metrics.go
```diff
@@ -0,0 +1,56 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+const (
+	cacheMetricsCollectionPeriod = 1 * time.Minute
+)
+
+func (f *sharedCacheFactory) collectMetrics() sharedCacheFactoryMetrics {
+	f.lock.RLock()
+	defer f.lock.RUnlock()
```
bigkevmcd, line 17: I meant to add, I suspect this should be a `.Lock()` not a `.RLock()`

https://go.dev/ref/spec#For_range

> If a map entry is created during iteration, that entry may be produced during the iteration or may be skipped.

pkg/cache/metrics.go
```diff
@@ -0,0 +1,56 @@
...
+
+func (f *sharedCacheFactory) startMetricsCollection(ctx context.Context) {
+	go func() {
+		contextID := metrics.ContextID(ctx)
+		timer := time.NewTimer(cacheMetricsCollectionPeriod)
+		defer timer.Stop()
+		for {
+			factoryMetrics := f.collectMetrics()
+			for gvk, count := range factoryMetrics.gvks {
+				metrics.IncTotalCachedObjects(contextID, gvk.Group, gvk.Version, gvk.Kind, float64(count))
```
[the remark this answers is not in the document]
aruiz14: I was trying to keep things as similar to how they were as possible, but since I like those suggestions and I'm already modifying the signature anyways, I'll change it, thanks.

pkg/cache/metrics.go
aruiz14: The lock protects `f.caches`, not the map being created. Given it was already protected, I only changed it to a `RWMutex`. It's unlikely to happen, but this may panic if a new cache is added/started so `f.caches` gets a write while reading here

moio (approved):
LGTM. As @bigkevmcd remarked, it would be best to have some kind of testing around this.

pkg/cache/metrics_test.go
```diff
@@ -0,0 +1,96 @@
...
+		t.Fatal(err)
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	t.Cleanup(cancel)
+
+	ctx = metrics.WithContextID(ctx, "test-ctx")
+
+	// Start metrics collection:
+	// It will report 0 items for every kind, since indexers is registered byt not started, which is enough for testing
```
bigkevmcd, line 74:
```suggestion
	// It will report 0 items for every kind, since indexers is registered but not started, which is enough for testing
```

pkg/cache/metrics_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+	cf := setupMockSharedClientFactory(t, gvr, gvk)
+	scf := NewSharedCachedFactory(cf, &SharedCacheFactoryOptions{MetricsCollectionPeriod: time.Millisecond * 200}).(*sharedCacheFactory)
+	_, err := scf.ForKind(gvk)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	t.Cleanup(cancel)
```
bigkevmcd, line 69: You're calling `cancel`  later in the code?

pkg/cache/metrics_test.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/prometheus/client_golang/prometheus/testutil"
+	"github.com/rancher/lasso/pkg/client"
+	"github.com/rancher/lasso/pkg/metrics"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+)
+
+func setupMockSharedClientFactory(t *testing.T, gvr schema.GroupVersionResource, gvk schema.GroupVersionKind) client.SharedClientFactory {
```
bigkevmcd, line 21: You could add another gvk/r to the test to make use of this setup logic?

pkg/cache/metrics_test.go
aruiz14: Yes, cancelling the same context twice is harmless, and eliminates the chance of the first assertion failing, so that `t.Fatal` will make the `cancel()` unreachable.
Also, it kind of hurts my eyes not seeing `defer cancel()` after creating the context 😅 I used `t.Cleanup` though to express it's actual cleanup.

bigkevmcd (dismissed):

moio (dismissed):

moio (approved):

moio (approved):

pkg/cache/metrics_test.go
```diff
@@ -0,0 +1,97 @@
...
+		t.Fatal(err)
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	t.Cleanup(cancel)
+
+	ctx = metrics.WithContextID(ctx, "test-ctx")
+
+	// Start metrics collection:
+	// It will report 0 items for every kind, since indexers is registered but not started, which is enough for testing
```
tomleb, line 74: I think it would be worth it to verify that the total changes based on how many are in the store/indexer. Would you mind explaining why you don't think that's necessary?

pkg/cache/metrics.go
```diff
@@ -0,0 +1,64 @@
...
+	defaultCacheMetricsCollectionPeriod = 1 * time.Minute
+)
+
+func (f *sharedCacheFactory) collectMetrics() sharedCacheFactoryMetrics {
+	f.lock.RLock()
+	defer f.lock.RUnlock()
+
+	gvks := make(map[schema.GroupVersionKind]int)
+	for gvk, c := range f.caches {
+		items := c.GetStore().List()
```
tomleb, line 21: I'd be curious to know if `List()` can be slow here (with Vai) such that it would be worth it to do something like:
```go
f.lock.RLock()
caches := map[gvk]cache.SharedBlabla
for gvk, c := range f.caches {
caches[gvk] = c
}
f.lock.RUnlock()

// Then to the listing with the copied caches
```

pkg/cache/metrics_test.go
bigkevmcd: Just about here we could do something like this...

```go
    testCache, err := scf.ForResourceKind(corev1.SchemeGroupVersion.WithResource("configmaps"), "ConfigMap", true)
    if err != nil {
        t.Fatal(err)
    }
    testConfigMap := &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: "testing", Namespace: "default"}, TypeMeta: metav1.TypeMeta{Kind: "ConfigMap", APIVersion: "v1"}}
    if err := testCache.GetStore().Add(testConfigMap); err != nil {
        t.Fatal(err)
    }   
```

Which would result in the metric count being `1` for ConfigMaps.

This would test that we're recording the number of resources in the cache correctly...

You could then call `testCache.GetStore().Delete(testConfigMap)` and verify that the count goes back to zero before cancelling the context and triggering the removal of all metrics?

pkg/cache/metrics_test.go
```diff
@@ -0,0 +1,97 @@
...
+	reg := prometheus.NewPedanticRegistry()
+	metrics.MustRegister(reg)
+	scf.startMetricsCollection(ctx)
+
+	time.Sleep(time.Second)
```
bigkevmcd, line 79: Also, as you're reducing the collection period to 200 Milliseconds, you could reduce the `time.Second` to `500 * time.Millisecond`, maybe even `300`?

bigkevmcd (changes requested):

pkg/cache/metrics.go
aruiz14: Good point, thanks! I've followed a similar approach using `maps.Clone` to get a local copy of the map.

pkg/cache/metrics_test.go
aruiz14: Good point, thanks, added!

bigkevmcd (approved):
Thanks for the changes 😄

tomleb (approved):

pkg/cache/metrics.go
tomleb: I didn't know about maps.Clone, nice.

aruiz14: Yeah, both [`maps`](https://pkg.go.dev/maps) and [`slices`](https://pkg.go.dev/slices) were (partially) promoted from their [experimental](https://pkg.go.dev/golang.org/x/exp) versions in recent versions of Go, which came along with the support for generics.

aruiz14:
@tomleb merging is restricted, could you please push the button whenever it suits you best? Thanks!

PR #99, opened 2024-08-30
sql: allow multiple informers to start in parallel 

moio:
Creating an Informer used to block creation of any other Informers for the whole duration of the warm-up time via one mutex.

Warm-up can be long if very many resources are involved, and the behavior blocked any unrelated Steve API calls (eg. UI).

This patch allows for multiple Informers to be created in parallel, with one mutex per gvk.

One overall `RWMutex` is still kept to avoid races with `Reset()`.

_this contains commits from https://github.com/rancher/lasso/pull/97 and https://github.com/rancher/lasso/pull/98, which should be merged first_


pkg/cache/sql/informer/factory/informer_factory.go
```diff
@@ -25,15 +27,16 @@ const EncryptAllEnvVar = "CATTLE_ENCRYPT_CACHE_ALL"
...
+	dbClient    DBClient
+	stopCh      chan struct{}
+	stopChMutex sync.RWMutex
+	encryptAll  bool
 
 	newInformer newInformer
 
-	cache map[schema.GroupVersionKind]*informer.Informer
+	informers       sync.Map
+	informerMutexes sync.Map
```
tomleb, line 39: Curious as to why the use of `sync.Map` instead of a simple map+mutex. There's a bit of added complexity with map with the different methods than a regular map, the fact that it's basically a `map[any]any` so needs casting everywhere and it's not clear when looking at it what the type of key or value is.

Also feels like it would be easier to maintain using a single map instead of two that needs to be kept in sync. Something like `map[gvk]*guardedInformer` where the `guardedInformer` is just a mtuex+the informer. (If you can come up with a better name, even better :D )

WDYT?

pkg/cache/sql/informer/factory/informer_factory.go
```diff
@@ -20,17 +21,22 @@ import (
...
 type CacheFactory struct {
-	informerCreateLock sync.RWMutex
-	wg                 wait.Group
-	dbClient           DBClient
-	stopCh             chan struct{}
-	encryptAll         bool
+	wg          wait.Group
+	dbClient    DBClient
+	stopCh      chan struct{}
+	stopChMutex sync.RWMutex
```
tomleb, line 33: nit: That mutex doesn't only guard the stopCh channel but also other fields like informers/informerMutexes. Seems like the name should be changed but I'm out of inspiration. Maybe just `lock`, with a comment saying what it is used for? WDYT?

pkg/cache/sql/informer/factory/informer_factory.go
```diff
@@ -20,17 +21,22 @@ import (
...
+const EncryptAllEnvVar = "CATTLE_ENCRYPT_CACHE_ALL"
+
 // CacheFactory builds Informer instances and keeps a cache of instances it created
 type CacheFactory struct {
-	informerCreateLock sync.RWMutex
-	wg                 wait.Group
-	dbClient           DBClient
-	stopCh             chan struct{}
-	encryptAll         bool
+	wg          wait.Group
```
tomleb, line 30: Doesn't have much to do with this PR but we're never calling `wg.Wait()`. We should probably be calling it after closing the `stopCh` channel... What do you think about fixing this in this PR? Otherwise I can create an issue and leave it for another.

tomleb (commented):
That's a nice addition. Some comments/requests for changes.

pkg/cache/sql/informer/factory/informer_factory.go
moio: Fair point.

I like `mutex` better so I'll go down that path unless an illumination comes in the form of a much better name. Thanks

pkg/cache/sql/informer/factory/informer_factory.go
tomleb: 1. Yeah I can see how having to deal with two maps and two locks would end up not very readable. I wonder if a similar approach would have worked with a single lock and two maps. Basically, lock the maps with single lock, initialize both the RWMutex and the Informer if it doesn't exists yet, unlock the maps with single lock, lock the per-type lock, start the informer or something.. Anyway, hard to reason with all these locks.


2. Yeah, if you're opened to `map[gvk]*guardedInformer + Mutex` and it works for our use case then we should go with that, I think it's going to be more readable.

moio: I tried but failed to come up with a satisfying solution with one only map of `guardedInformer`s.

The reason for that is management of the time interval when:

- an informer for a given GVK has not been created yet (so a `guardedInformer.informer` is still nil) but
- the indexer-specific mutex has to be locked, to prevent multiple Informers may start at the same time for that GVK (so `guardedInformer.mutex` has to be non-nil)

Code-wise that corresponds to these lines: https://github.com/moio/lasso/blob/075de697fc1615403d6ca6d3f982a743673d3fca/pkg/cache/sql/informer/factory/informer_factory.go#L102-L117

Having both the informer and the informer mutex in the same map means I have to lock/unlock the map twice: first get the mutex (and set it if it does not exist - bringing in logic similar to [LoadOrStore](https://cs.opensource.google/go/go/+/refs/tags/go1.23.1:src/sync/map.go;l=227), which is not trivial), then to set the informer once it's created.

moio: To make it a bit more concrete, this is what I came up with:

https://github.com/moio/lasso/pull/1/files

To me that is subjectively harder to follow despite types, mainly because I am breaking the "Lock then defer" pattern.

The other problem it has is that [golang's RWMutex does not support lock upgrading](https://github.com/golang/go/issues/38859#issuecomment-623742366), which, IIUC, means that code has a risk of deadlocking if two goroutines enter [this line](https://github.com/moio/lasso/pull/1/files#diff-30ea1a028c2e55261cf0c7fd16267f209c53187a2fa4372afc2580001b9a668bR114).

Happy to hear your thoughts.

pkg/cache/sql/informer/factory/informer_factory.go
moio: fixed in 81eb208

pkg/cache/sql/informer/factory/informer_factory.go
moio: fixed in be00af8

pkg/cache/sql/informer/factory/informer_factory.go
tomleb: How about something like this? [pkg/cache/sql/informer/factory/informer_factory.go#L95-L129](https://github.com/rancher/lasso/blob/913ec0e0ca62faed914029b9cb5178f7ccdcb06f/pkg/cache/sql/informer/factory/informer_factory.go#L95-L129)

My assumption (and let me know if I'm wrong) is that there's two things that takes time:
1. newInformer: should be pretty quick iirc as it's only creating a DB and setting up the tables
2. waiting for caches to sync: this one I expect to take the longest when there's a high number of objects

So with these assumptions, the code locks the global mutex for a very short amount of time, letting other GVK start at the same time. Then, each GVK is locked until informer is set and started (should be quick).

Finally, no locks are taken while we wait for the caches to sync.

Let me know your thoughts.

moio: I like your proposal - I was overcomplicating my life because `RWMutex`es, while a plain `Mutex` is probably as effective, and much more readable.

One problem with the code was `gi.informer` was never set - fixed now.

Another thing that's entirely opinable is the placement of `WaitForCacheSync`. That method spinlocks until the informer is ready, so in principle it can be outside of the lock as you put it. Nevertheless there is no real advantage in having it called by non-creating goroutines, so I moved it inside the lock. That had the nice side effect of allowing to use defer on the `gi.mutex`, which IMO adds a bit of readability.

Concluding this discussion and waiting for your next round of comments on the patch overall.

tomleb: >  Nevertheless there is no real advantage in having it called by non-creating goroutines, so I moved it inside the lock.

That's perfect :+1: it's better this way

pkg/cache/sql/informer/factory/informer_factory_test.go
```diff
@@ -63,7 +63,7 @@ func TestCacheFor(t *testing.T) {
 		fields := [][]string{{"something"}}
 		expectedGVK := schema.GroupVersionKind{}
 		sii := NewMockSharedIndexInformer(gomock.NewController(t))
-		sii.EXPECT().HasSynced().Return(true)
+		sii.EXPECT().HasSynced().Return(true).Times(2)
```
tomleb, line 66: Do we care that HasSynced has been called twice here? And are we sure that it will always be called twice? I just want to avoid flaky tests.

tomleb (commented):
One comment on test, otherwise looks good. thanks!

pkg/cache/sql/informer/factory/informer_factory_test.go
moio: ack, fixed in 4f36817

tomleb (approved):

aruiz14 (approved):

MbolotSuse (approved):
Approving to unblock merging (since the required 2 approvals have been made).

PR #100, opened 2024-09-05
Switch from github.com/golang/mock to go.uber.org/mock

aruiz14:
## Issue: rancher/rancher#46972

- Adds missing `go:generate`, so all code can now be generated running `go generate ./...`
- Refresh generated files
- Update imports to the new dependency.
- `go mod tidy`


ericpromislow (approved):
Looks fine. I especially like the way the uber version writes out the command in generated files that was used to generate them.

aruiz14:
Thanks @ericpromislow, would you mind also taking a look at the equivalent PR in [wrangler](https://github.com/rancher/wrangler/pull/468)?

moio (dismissed):
Switching to a maintained version is a good thing (tm)

approved

moio (approved):

ericpromislow (approved):
Good after this last change

PR #101, opened 2024-09-05
Integration tests

MbolotSuse:
Related to rancher/rancher#45243. Adds some env-test based tests that attempt to use the SQL cache closer to how it's used in production. This only covers some basic filter cases, but sets up the framework to add tests for things like: transform functions, sorting, and pagination.

Note: This PR may need to be re-opened on a branch in the repo to allow the CI to run successfully.


tomleb (approved):
LGTM, nice work :tada:

scripts/test.sh
```diff
@@ -1,3 +1,9 @@
 #!/bin/bash
 
+if ! command -v setup-envtest; then
+	echo "setup-env is required for tests"
+	exit -1
+fi
+
+export KUBEBUILDER_ASSETS=$(setup-envtest use -p path)
```
tomleb, line 8:
```suggestion
minor=$(go list -m all | grep k8s.io/client-go | cut -d' ' -f2 | cut -d. -f2)
version="1.$minor.x"
export KUBEBUILDER_ASSETS=$(setup-envtest use -p path "$version")
```

That way we test for a specific version so that in the future when an older branch runs the test it doesn't test with an unsupported version of Rancher. WDYT?

pkg/cache/sql/integration_test.go
```diff
@@ -0,0 +1,324 @@
...
+	testEnv   *envtest.Environment
+	clientset kubernetes.Clientset
+	restCfg   rest.Config
+}
+
+func (i *IntegrationSuite) SetupSuite() {
+	i.testEnv = &envtest.Environment{}
+	restCfg, err := i.testEnv.Start()
+	i.Require().NoError(err)
+	i.Require().NotNil(restCfg)
```
bigkevmcd, line 41: nit: I don't think you need to test for this being non-nil, worst case, it'll cause an error just a couple of lines below.

pkg/cache/sql/integration_test.go
```diff
@@ -0,0 +1,324 @@
...
+		i.Run(test.name, func() {
+			options := informer.ListOptions{
+				Filters: test.filters,
+			}
+			partitions := []partition.Partition{defaultPartition}
+			ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+			cfgMaps, total, continueToken, err := cache.ListByOptions(ctx, options, partitions, testNamespace)
+			i.Require().NoError(err)
+			// cancel here so that we don't build up a large list of defer'd cancels
+			cancel()
```
bigkevmcd, line 245: nit: I think deferring the cancel would be ok here, because it would fire when leaving the anonymous function?

scripts/test.sh
```diff
@@ -1,3 +1,9 @@
 #!/bin/bash
```
maxsokolovsky, line 1: I know you weren't the one who created the file, but do you think we should make this script executable as we do with scripts in other repos?

scripts/test.sh
MbolotSuse: I'm not sure about that. Per the versioning policy for both wrangler and lasso( [lasso](https://github.com/rancher/rancher/issues/46901), [wrangler](https://github.com/rancher/rancher/issues/46900)), this should support a range of k8s versions, not a single version as determined by client-go. So if we were going to do something like this, ideally this should iterate over a range of versions, run setup with the version, and then run the test suite. However, since the versioning system isn't integrated yet, I'm not sure that this would be something that we want to integrate.

I think what you proposed here is probably a bit more accurate, so I can implement it if we want. However, I think that the lifespan is limited an eventually we will want to remove the go mod scraping in favor of the approach that I proposed above, so I'm not sure that it's worth adding right now. WDYT?

pkg/cache/sql/integration_test.go
```diff
@@ -0,0 +1,324 @@
...
+	defer cacheFactory.Reset()
+
+	// doesn't match the filter for somekey == somevalue
+	notMatches := configMapWithAnnotations("not-matches-filter", map[string]string{"somekey": "notequal"})
+	// has no annotations, shouldn't match any filter
+	missing := configMapWithAnnotations("missing", nil)
+	createConfigMaps(notMatches, missing)
+
+	configMapNames := []string{matches.Name, partialMatches.Name, notMatches.Name, missing.Name}
+	i.waitForCacheReady(configMapNames, testNamespace, cache)
```
maxsokolovsky, line 103:
```suggestion
	err = i.waitForCacheReady(configMapNames, testNamespace, cache)
```

pkg/cache/sql/integration_test.go
```diff
@@ -0,0 +1,324 @@
...
+			cfgMaps, total, continueToken, err := cache.ListByOptions(ctx, options, partitions, testNamespace)
+			i.Require().NoError(err)
+			// cancel here so that we don't build up a large list of defer'd cancels
+			cancel()
+			i.Require().Equal(len(test.wantNames), total)
+			// since there's no additional pages, the continue token should be empty
+			i.Require().Equal("", continueToken)
+			i.Require().NotNil(cfgMaps)
+			i.Require().Len(cfgMaps.Items, len(test.wantNames))
+			requireNames := map[string]struct{}{}
```
bigkevmcd, line 251: Maybe this could be simplified by using set https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#Set.Equal ?

Create a set from the want names, and one from the cfgMaps.Items names and compare them?

Alternatively, you could build a slice of the names from the cfgMaps.Items names and compare them?

It would make it clearer what it's doing?

pkg/cache/sql/integration_test.go
```diff
@@ -0,0 +1,324 @@
...
+}
+
+func (i *IntegrationSuite) createCacheAndFactory(fields [][]string, transformFunc cache.TransformFunc) (*factory.Cache, *factory.CacheFactory, error) {
+	cacheFactory, err := factory.NewCacheFactory()
+	if err != nil {
+		return nil, nil, fmt.Errorf("unable to make factory: %w", err)
+	}
+	dynamicClient, err := dynamic.NewForConfig(&i.restCfg)
+	if err != nil {
+		return nil, nil, fmt.Errorf("unable to make dynamicClient: %w", err)
```
bigkevmcd, line 270: Instead of returning an error maybe this fails the test?

scripts/test.sh
```diff
@@ -1,3 +1,9 @@
...
 
+if ! command -v setup-envtest; then
```
bigkevmcd, line 3: Maybe this could say how to install it?

bigkevmcd (commented):
Tests are a good start.

scripts/test.sh
tomleb: I also considered running tests against multiple supported version but figured that would be too much work for a first try, we don't have to get everything right the first time. That said I still think using the minor from client-go is a step in the right direction and it will also prevent CI breakage (eg: "yesterday the CI worked but today it doesn't because upstream released a new version unrelated to this PR's change")

pkg/cache/sql/integration_test.go
```diff
@@ -0,0 +1,324 @@
...
+type IntegrationSuite struct {
+	suite.Suite
+	testEnv   *envtest.Environment
+	clientset kubernetes.Clientset
+	restCfg   rest.Config
+}
+
+func (i *IntegrationSuite) SetupSuite() {
+	i.testEnv = &envtest.Environment{}
+	restCfg, err := i.testEnv.Start()
```
maxsokolovsky, line 39: I think it'd be good to check for the presence of the `KUBEBUILDER_ASSETS` environment variable early and fail if it's missing. Otherwise, the error message is confusing ("etcd is missing from kubebuilder binaries").

pkg/cache/sql/integration_test.go
MbolotSuse: For side functions like this I prefer to return an error. The main function that's actually running the test can make the require/fail assertions to actually fail the test.

scripts/test.sh
MbolotSuse: Alright, on the basis of reducing flakiness I will add it.

maxsokolovsky (changes requested):

scripts/test.sh
```diff
@@ -1,3 +1,9 @@
...
+	echo "setup-env is required for tests"
+	exit -1
```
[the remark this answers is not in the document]
MbolotSuse: Bash manual: https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html, seems to indicate that 127 (command not found) is the most appropriate for this case. Any disagreement?

scripts/test.sh
MbolotSuse: Generally speaking I'd prefer to limit the changes here in favor of getting the code in. I can make this one change, but outside of this please limit the asks to the PR's functionality.

pkg/cache/sql/integration_test.go
MbolotSuse: I disagree here. The test script does some checking here to make sure that things are properly setup. If you choose to use something else, I think that's on you (as the developer) to debug. If we start integrity checking this, it's tough for us to choose when to stop (i.e. what if someone just sets this to a blank string? What if they set filepaths to random places that don't include the binaries? What if the binaries are there but they aren't actually the real programs?).

bigkevmcd: KUBEBUILDER_ASSETS is definitive https://github.com/kubernetes-sigs/controller-runtime/blob/38546806f2faf5973e3321a7bd5bb3afdbb5767d/pkg/internal/testing/process/bin_path_finder.go#L58-L61

But other options are possible, so it definitely seems reasonable to not check for it.

maxsokolovsky: It's for developer experience - if I just click and run this test suite, I won't understand the error message. I wouldn't know that I have to run this test in a special way - via this script, and that script also runs other tests, which I may not want.

MbolotSuse: Right, which is on you as the developer to figure out. If you'd like I can add a comment about the requirement to the file so it's immediately apparent on failure. However, I can't agree on adding checks for this case.

tomleb: A README or a section in the README could help here. You could:
- Point the user to it when testEnv.Start() fails
- Point the user to it when `setup-envtest` is not installed (in the scripts/test.sh file, not in the test)

See https://github.com/rancher/rancher/tree/main/tests/controllers for an example in r/r.

pkg/cache/sql/integration_test.go
MbolotSuse: I've changed this to use sets (and I changed the other place which was doing the same thing).

scripts/test.sh
MbolotSuse: I've referred this over to the readme like the other tests, let me know if you were thinking of something different.

pkg/cache/sql/integration_test.go
MbolotSuse: I've added information to the readme and pointed the user to it in the error message.

pkg/cache/sql/integration_test.go
```diff
@@ -0,0 +1,322 @@
...
+	suite.Suite
+	testEnv   envtest.Environment
+	clientset kubernetes.Clientset
+	restCfg   rest.Config
+}
+
+func (i *IntegrationSuite) SetupSuite() {
+	i.testEnv = envtest.Environment{}
+	restCfg, err := i.testEnv.Start()
+	i.Require().NoError(err, "error when staring env test - this is likely because setup-envtest wasn't done. Check the README for more information")
```
tomleb, line 41:
```suggestion
	i.Require().NoError(err, "error when starting env test - this is likely because setup-envtest wasn't done. Check the README for more information")
```

maxsokolovsky (approved):

tomleb (approved):

moio (approved):
Excellent work, thanks for taking the lead on this @MbolotSuse !

Very appreciated.

PR #108, opened 2024-09-18
sql: remove obsolete options

moio:
This is to remove options that only make sense when in-memory SQLite was an option.


tomleb (approved):

aruiz14 (approved):

MbolotSuse (approved):
Approving due to the required 2 approvals being present.

PR #111, opened 2024-09-24
sql: silence informer errors on unwatchable resources

moio:
Listable-but-unwatchable resources such as `nodes metrics.k8s.io/v1beta1` allow an Informer to be started, but it will by default log all watch failures.

Add a boolean flag to silence those errors.

partially addresses https://github.com/rancher/rancher/issues/46030
supports https://github.com/rancher/steve/pull/284

### DIFF
--- a/pkg/cache/sql/informer/factory/informer_factory.go
+++ b/pkg/cache/sql/informer/factory/informer_factory.go
@@ -14,6 +14,7 @@ import (
 	"github.com/rancher/lasso/pkg/cache/sql/informer"
 	sqlStore "github.com/rancher/lasso/pkg/cache/sql/store"
 	"github.com/rancher/lasso/pkg/log"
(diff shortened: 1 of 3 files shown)


ericpromislow:
Looks ok, but how is the calling code in `steve` supposed to know which resources are watchable and which aren't? I assume there's something in one of the schema packages, but haven't found it.

ericpromislow (commented):
Looks ok, but I'd like to know how clients determine when `watchable` should be true.

moio:
Yes, we have a method for that in Steve https://github.com/rancher/steve/blob/ea177193f12ade5f64ac4ef58354a629329da9e2/pkg/controllers/schema/schemas.go#L111

This PR actually supports a sister Steve PR https://github.com/rancher/steve/pull/284

ericpromislow (approved):
Looks good

PR #112, opened 2024-09-24
Wrap the main and count queries in ListByOptions with a transaction

maxsokolovsky:
This ensures that the main query and the conditionally run subsequent count query are within the same transaction context.
This will make it so that the second query that counts the number of rows returned by the first will always see the actual number, without the possibility of modifications to the data introduced in between query executions.


maxsokolovsky:
@moio, I've had to re-generate the same interface as a mock in the package where I have the main test. Looks like there are a few instances where a single interface has multiple mock implementations. Are they all needed? Or do we do this just because Go doesn't allow to import exported types defined in **test** code? If so, could we extract mocks to a non-test package?

moio:
I share the feeling current code is using a lot of almost-duplicated interfaces and mocking, and that at times confuses me as well.

I value any step big or small to make interfaces and tests more understandable.

About extracting mocks to a non-test package: I have no problems with that, I am not aware of side effects other than not being immediately clear from the filename that the code in question is test code. I am completely OK if that message is conveyed to readers in some other way, eg. with type and function names.

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -376,15 +385,21 @@ func (l *ListOptionIndexer) ListByOptions(ctx context.Context, lo ListOptions, p
 	if limit > 0 || offset > 0 {
 		countStmt := l.Prepare(countQuery)
 		defer l.CloseStmt(countStmt)
-		rows, err = l.QueryForRows(ctx, countStmt, countParams...)
+		txStmt := tx.Stmt(countStmt)
+		rows, err := txStmt.QueryContext(ctx, countParams...)
 		if err != nil {
+			_ = tx.Cancel()
```
moio, line 391: Being a bit picky: Cancel (actually Rollback) might return an error different from the query error, and we might want to expose both.

Ideally capture the err here, and make sure it is wrapped into the returned error if possible.

Thanks!

moio (approved):
LGTM. In future, we might want to extract the common code for canceling transactions and handling errors in a helper/utility function, but I do not see it as necessary before this can be merged.

ericpromislow (changes requested):
Wondering if we can just call `tx.Rollback()` after a group of queries, as there's nothing to change in the database.

moio:
@ericpromislow we can - after a group of read-only queries commit and rollback are synonymous.

Are you suggesting any change in this PR?

prachidamle:
This PR is part of https://github.com/rancher/rancher/issues/46410

ericpromislow (approved):
Let's get this change in and later change useless `Commit`s to `Rollback`s after only reads

PR #116, opened 2024-11-01
Fix the file permissions on the sqlite db.

bigkevmcd:
This ensures that the permissions on the file that SQLite is using are correct (0600) i.e. user read/write.

Part of https://github.com/rancher/rancher/issues/47825

### DIFF
--- a/pkg/cache/sql/db/client.go
+++ b/pkg/cache/sql/db/client.go
@@ -324,11 +324,8 @@ func (c *Client) NewConnection() error {
 		return err
 	}
 
-	// Touch the file first which means that when SQLite opens it, it will
-	// retain the permissions.
-	// This is done because we can't control the permissions via the sql.Open
-	// call and when the file gets created is down to when sqlite persists
-	// something.
+	// Set the permissions in advance, because we can't control them if
+	// the file is created by a sql.Open call instead.
 	if err := touchFile(InformerObjectCacheDBPath, informerObjectCachePerms); err != nil {
 		return nil
 	}
--- a/pkg/cache/sql/db/client.go
+++ b/pkg/cache/sql/db/client.go
@@ -24,7 +24,7 @@ const (
 	// InformerObjectCacheDBPath is where SQLite's object database file will be stored relative to process running lasso
 	InformerObjectCacheDBPath = "informer_object_cache.db"
 
-	informerObjectCachePerms fs.FileMode = 0600
+	informerObjectCachePerms fs.FileMode = 0o600
 )
 
 // Client is a database client that provides encrypting, decrypting, and database resetting.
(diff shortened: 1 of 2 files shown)


bigkevmcd:
This fixes the primary object cache, but I'm not sure where the other resources are stored.

pkg/cache/sql/db/client.go
```diff
@@ -22,6 +23,8 @@ import (
 const (
 	// InformerObjectCacheDBPath is where SQLite's object database file will be stored relative to process running lasso
 	InformerObjectCacheDBPath = "informer_object_cache.db"
+
+	informerObjectCachePerms fs.FileMode = 0600
```
pjbgf, line 27: nit, we could make it explicit that this is a literal octal:

```suggestion
	informerObjectCachePerms fs.FileMode = 0o600
```

pkg/cache/sql/db/client.go
```diff
@@ -343,3 +353,14 @@ func (c *Client) NewConnection() error {
...
 }
+
+// This acts like "touch" for non-existent files...setting the correct
+// permissions.
+func touchFile(filename string, perms fs.FileMode) error {
+	f, err := os.OpenFile(filename, os.O_RDWR|os.O_CREATE, perms)
+	if err != nil {
+		return err
+	}
+
```
pjbgf, line 364: If the file already existed, this won't enforce the target file mode. We could add a `chmod` here to that effect. That way, we always ensure the correct file mode is in place. Thoughts?

bigkevmcd: I am not sure we _need_ to just now?

Definitely could be documented, but the only place it's called from is just after it's [deleted](https://github.com/rancher/lasso/blob/3c49d3df84e2025d5f879209b6254edca18c66f1/pkg/cache/sql/db/client.go#L322) ?

It could be a generic "test for file, create if doesn't exist with right perms otherwise, os.Chmod" function? Admittedly, that's racy, so we could opt to do that just after we create it...

Happy to add it in case it's used later for something else.

pjbgf (approved):

moio (dismissed):
lgtm

pkg/cache/sql/db/client.go
```diff
@@ -320,6 +324,15 @@ func (c *Client) NewConnection() error {
 		return err
 	}
 
+	// Touch the file first which means that when SQLite opens it, it will
+	// retain the permissions.
+	// This is done because we can't control the permissions via the sql.Open
```
ericpromislow, line 329: This comment could be clearer. Maybe something like:

Set the permissions in advance, because we can't control them if
the file is created by a `sql.Open` call instead.

ericpromislow (changes requested):
Fine, but I'd like the comment clarified.

moio (approved):

bigkevmcd:
@ericpromislow Can you give this a 👍🏻 ?

moio (approved):

ericpromislow (approved):
LG now

PR #117, opened 2024-11-06
Support database queries on arbitrary labels

ericpromislow:
Related to [#46333](https://github.com/rancher/rancher/issues/46333)

This PR needs to get merged first before I can submit the PR for Steve,
which caches the labels. If you prefer, I'll submit a PR with steve that
temporarily pulls in `../lasso` in the go.mod file.

Note that this is the search syntax I've implemented:

```
curl -sk https://HOSTANDPORT/v1/configmaps?filter=metadata.labels%5bLABELNAME%5d=LABELVALUE
```

If `LABELVALUE` is quoted, an exact match is made. Otherwise partial matching is done on the value,
like for other `A=B` queries.

Note that the exact `LABELNAME` must be specified after `metadata.labels`, between the escaped square brackets. 
This is similar to how you always have to provide the full name of a field to the left of `=`.

This PR supersedes the experimental PR #110

### DIFF
--- a/pkg/cache/sql/db/client.go
+++ b/pkg/cache/sql/db/client.go
@@ -10,8 +10,6 @@ import (
 	"database/sql"
 	"encoding/gob"
 	"fmt"
-	"github.com/sirupsen/logrus"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"io/fs"
 	"os"
 	"reflect"
--- a/pkg/cache/sql/db/client.go
+++ b/pkg/cache/sql/db/client.go
@@ -283,7 +283,7 @@ func (c *Client) Upsert(tx TXClient, stmt *sql.Stmt, key string, obj any, should
 func (c *Client) UpsertLabels(tx TXClient, stmt *sql.Stmt, key string, obj any, shouldEncrypt bool) error {
 	k8sObj, ok := obj.(*unstructured.Unstructured)
 	if !ok {
-		logrus.Errorf("QQQ: UpsertLabels: Error?: Can't convert obj into an unstructured thing.")
+		logrus.Debugf("UpsertLabels: Error?: Can't convert obj into an unstructured thing.")
 		return nil
 	}
 	incomingLabels := k8sObj.GetLabels()
--- a/pkg/cache/sql/db/client.go
+++ b/pkg/cache/sql/db/client.go
@@ -230,7 +230,7 @@ func (c *Client) ReadInt(rows Rows) (int, error) {
 // If forWriting is true, this method blocks until all other concurrent forWriting
 // transactions have either committed or rolled back.
 // If forWriting is false, it is assumed the returned transaction will exclusively
-// be used for DQL (eg. SELECT) queries.
+// be used for DQL (e.g. SELECT) queries.
 // Not respecting the above rule might result in transactions failing with unexpected
 // SQLITE_BUSY (5) errors (aka "Runtime error: database is locked").
 // See discussion in https://github.com/rancher/lasso/pull/98 for details
@@ -278,22 +278,6 @@ func (c *Client) Upsert(tx TXClient, stmt *sql.Stmt, key string, obj any, should
 	}
 
 	return tx.StmtExec(tx.Stmt(stmt), key, objBytes, dataNonce, kid)
-}
-
-func (c *Client) UpsertLabels(tx TXClient, stmt *sql.Stmt, key string, obj any, shouldEncrypt bool) error {
-	k8sObj, ok := obj.(*unstructured.Unstructured)
-	if !ok {
-		logrus.Debugf("UpsertLabels: Error?: Can't convert obj into an unstructured thing.")
-		return nil
-	}
-	incomingLabels := k8sObj.GetLabels()
-	for k, v := range incomingLabels {
-		err := tx.StmtExec(tx.Stmt(stmt), key, k, v)
-		if err != nil {
-			return err
-		}
-	}
-	return nil
 }
 
 // toBytes encodes an object to a byte slice
--- a/pkg/cache/sql/db/client.go
+++ b/pkg/cache/sql/db/client.go
@@ -10,6 +10,8 @@ import (
 	"database/sql"
 	"encoding/gob"
 	"fmt"
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"io/fs"
 	"os"
 	"reflect"
@@ -276,6 +278,22 @@ func (c *Client) Upsert(tx TXClient, stmt *sql.Stmt, key string, obj any, should
 	}
 
 	return tx.StmtExec(tx.Stmt(stmt), key, objBytes, dataNonce, kid)
+}
+
+func (c *Client) UpsertLabels(tx TXClient, stmt *sql.Stmt, key string, obj any, shouldEncrypt bool) error {
+	k8sObj, ok := obj.(*unstructured.Unstructured)
+	if !ok {
+		logrus.Errorf("QQQ: UpsertLabels: Error?: Can't convert obj into an unstructured thing.")
+		return nil
+	}
+	incomingLabels := k8sObj.GetLabels()
+	for k, v := range incomingLabels {
+		err := tx.StmtExec(tx.Stmt(stmt), key, k, v)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
 // toBytes encodes an object to a byte slice
--- a/pkg/cache/sql/informer/listoption_indexer.go
+++ b/pkg/cache/sql/informer/listoption_indexer.go
@@ -48,7 +48,7 @@ var (
 const (
 	matchFmt                 = `%%%s%%`
 	strictMatchFmt           = `%s`
-	escapeBackslashDirective = ` ESCAPE '\'` // The leading space is crucial for unit tests only
+	escapeBackslashDirective = ` ESCAPE '\'` // The leading space is crucial for unit tests only '
 	createFieldsTableFmt     = `CREATE TABLE "%s_fields" (
 			key TEXT NOT NULL PRIMARY KEY,
             %s
@@ -270,79 +270,17 @@ type QueryInfo struct {
 //   - a continue token, if there are more pages after the returned one
 //   - an error instead of all of the above if anything went wrong
 func (l *ListOptionIndexer) ListByOptions(ctx context.Context, lo ListOptions, partitions []partition.Partition, namespace string) (*unstructured.UnstructuredList, int, string, error) {
-	queryInfo, err := l.constructQuery(lo, partitions, namespace)
+	queryInfo, err := l.constructQuery(lo, partitions, namespace, db.Sanitize(l.GetName()))
 	if err != nil {
 		return nil, 0, "", err
 	}
-	// execute
-	query := queryInfo.query
-	stmt := l.Prepare(query)
-	defer l.CloseStmt(stmt)
-
-	tx, err := l.BeginTx(ctx, false)
-	if err != nil {
-		return nil, 0, "", err
-	}
-
-	params := queryInfo.params
-	txStmt := tx.Stmt(stmt)
-	rows, err := txStmt.QueryContext(ctx, params...)
-	if err != nil {
-		if cerr := tx.Cancel(); cerr != nil {
-			return nil, 0, "", fmt.Errorf("failed to cancel transaction (%v) after error: %w", cerr, err)
-		}
-		return nil, 0, "", &db.QueryError{QueryString: query, Err: err}
-	}
-	items, err := l.ReadObjects(rows, l.GetType(), l.GetShouldEncrypt())
-	if err != nil {
-		if cerr := tx.Cancel(); cerr != nil {
-			return nil, 0, "", fmt.Errorf("failed to cancel transaction (%v) after error: %w", cerr, err)
-		}
-		return nil, 0, "", err
-	}
-
-	total := len(items)
-	countQuery := queryInfo.countQuery
-	countParams := queryInfo.countParams
-	limit := queryInfo.limit
-	offset := queryInfo.offset
-	// if limit or offset were set, execute counting of all rows
-	if limit > 0 || offset > 0 {
-		countStmt := l.Prepare(countQuery)
-		defer l.CloseStmt(countStmt)
-		txStmt := tx.Stmt(countStmt)
-		rows, err := txStmt.QueryContext(ctx, countParams...)
-		if err != nil {
-			if cerr := tx.Cancel(); cerr != nil {
-				return nil, 0, "", fmt.Errorf("failed to cancel transaction (%v) after error: %w", cerr, err)
-			}
-			return nil, 0, "", fmt.Errorf("error executing query: %w", err)
-		}
-		total, err = l.ReadInt(rows)
-		if err != nil {
-			if cerr := tx.Cancel(); cerr != nil {
-				return nil, 0, "", fmt.Errorf("failed to cancel transaction (%v) after error: %w", cerr, err)
-			}
-			return nil, 0, "", fmt.Errorf("error reading query results: %w", err)
-		}
-	}
-	if err := tx.Commit(); err != nil {
-		return nil, 0, "", err
-	}
-
-	continueToken := ""
-	if limit > 0 && offset+len(items) < total {
-		continueToken = fmt.Sprintf("%d", offset+limit)
-	}
-
-	return toUnstructuredList(items), total, continueToken, nil
+	return l.executeQuery(ctx, queryInfo)
 }
 
-func (l *ListOptionIndexer) constructQuery(lo ListOptions, partitions []partition.Partition, namespace string) (*QueryInfo, error) {
+func (l *ListOptionIndexer) constructQuery(lo ListOptions, partitions []partition.Partition, namespace string, dbName string) (*QueryInfo, error) {
 	queryInfo := &QueryInfo{}
 
 	// 1- First, what kind of filtering will be doing?
-	dbName := db.Sanitize(l.GetName())
 	// 1.1- Intro: SELECT and JOIN clauses
 	query := fmt.Sprintf(`SELECT o.object, o.objectnonce, o.dekid FROM "%s" o`, dbName)
 	query += "\n  "
@@ -506,6 +444,10 @@ func (l *ListOptionIndexer) constructQuery(lo ListOptions, partitions []partitio
 		offsetClause = "\n  OFFSET ?"
 		params = append(params, offset)
 	}
+	if limit == 0 && offset == 0 {
+		countQuery = ""
+		countParams = []any{}
+	}
 
 	// assemble and log the final query
 	query += limitClause
@@ -521,6 +463,70 @@ func (l *ListOptionIndexer) constructQuery(lo ListOptions, partitions []partitio
 		offset:      offset,
 	}
 	return queryInfo, nil
+}
+
+func (l *ListOptionIndexer) executeQuery(ctx context.Context, queryInfo *QueryInfo) (*unstructured.UnstructuredList, int, string, error) {
+	// execute
+	query := queryInfo.query
+	stmt := l.Prepare(query)
+	defer l.CloseStmt(stmt)
+
+	tx, err := l.BeginTx(ctx, false)
+	if err != nil {
+		return nil, 0, "", err
+	}
+
+	params := queryInfo.params
+	txStmt := tx.Stmt(stmt)
+	rows, err := txStmt.QueryContext(ctx, params...)
+	if err != nil {
+		if cerr := tx.Cancel(); cerr != nil {
+			return nil, 0, "", fmt.Errorf("failed to cancel transaction (%v) after error: %w", cerr, err)
+		}
+		return nil, 0, "", &db.QueryError{QueryString: query, Err: err}
+	}
+	items, err := l.ReadObjects(rows, l.GetType(), l.GetShouldEncrypt())
+	if err != nil {
+		if cerr := tx.Cancel(); cerr != nil {
+			return nil, 0, "", fmt.Errorf("failed to cancel transaction (%v) after error: %w", cerr, err)
+		}
+		return nil, 0, "", err
+	}
+
+	total := len(items)
+	countQuery := queryInfo.countQuery
+	countParams := queryInfo.countParams
+	limit := queryInfo.limit
+	offset := queryInfo.offset
+	if countQuery != "" {
+		countStmt := l.Prepare(countQuery)
+		defer l.CloseStmt(countStmt)
+		txStmt := tx.Stmt(countStmt)
+		rows, err := txStmt.QueryContext(ctx, countParams...)
+		if err != nil {
+			if cerr := tx.Cancel(); cerr != nil {
+				return nil, 0, "", fmt.Errorf("failed to cancel transaction (%v) after error: %w", cerr, err)
+			}
+			return nil, 0, "", fmt.Errorf("error executing query: %w", err)
+		}
+		total, err = l.ReadInt(rows)
+		if err != nil {
+			if cerr := tx.Cancel(); cerr != nil {
+				return nil, 0, "", fmt.Errorf("failed to cancel transaction (%v) after error: %w", cerr, err)
+			}
+			return nil, 0, "", fmt.Errorf("error reading query results: %w", err)
+		}
+	}
+	if err := tx.Commit(); err != nil {
+		return nil, 0, "", err
+	}
+
+	continueToken := ""
+	if limit > 0 && offset+len(items) < total {
+		continueToken = fmt.Sprintf("%d", offset+limit)
+	}
+
+	return toUnstructuredList(items), total, continueToken, nil
 }
 
 func (l *ListOptionIndexer) validateColumn(column string) error {
--- a/pkg/cache/sql/informer/listoption_indexer.go
+++ b/pkg/cache/sql/informer/listoption_indexer.go
@@ -442,30 +442,25 @@ func (l *ListOptionIndexer) constructQuery(lo ListOptions, partitions []partitio
 		offsetClause = "\n  OFFSET ?"
 		params = append(params, offset)
 	}
-	if limit == 0 && offset == 0 {
-		// We don't need to find the number of items, so clear these queries.
-		countQuery = ""
-		countParams = []any{}
+	if limit > 0 || offset > 0 {
+		query += limitClause
+		query += offsetClause
+		queryInfo.countQuery = countQuery
+		queryInfo.countParams = countParams
+		queryInfo.limit = limit
+		queryInfo.offset = offset
 	}
+	// Otherwise leave these as default values and the executor won't do pagination work
 
-	// assemble and log the final query
-	query += limitClause
-	query += offsetClause
 	logrus.Debugf("ListOptionIndexer prepared statement: %v", query)
 	logrus.Debugf("Params: %v", params)
-	queryInfo = &QueryInfo{
-		query:       query,
-		params:      params,
-		countQuery:  countQuery,
-		countParams: countParams,
-		limit:       limit,
-		offset:      offset,
-	}
+	queryInfo.query = query
+	queryInfo.params = params
+
 	return queryInfo, nil
 }
 
 func (l *ListOptionIndexer) executeQuery(ctx context.Context, queryInfo *QueryInfo) (*unstructured.UnstructuredList, int, string, error) {
-	// execute
 	stmt := l.Prepare(queryInfo.query)
 	defer l.CloseStmt(stmt)
 
--- a/pkg/cache/sql/informer/listoption_indexer.go
+++ b/pkg/cache/sql/informer/listoption_indexer.go
@@ -288,11 +288,7 @@ func (l *ListOptionIndexer) constructQuery(lo ListOptions, partitions []partitio
 	query += fmt.Sprintf(`JOIN "%s_fields" f ON o.key = f.key`, dbName)
 	if hasLabelFilter(lo.Filters) {
 		query += "\n  "
-		leftOuterPart := ""
-		if hasNegativeExistenceTest(lo.Filters) {
-			leftOuterPart = "LEFT OUTER "
-		}
-		query += fmt.Sprintf(`%sJOIN "%s_labels" lt ON o.key = lt.key`, leftOuterPart, dbName)
+		query += fmt.Sprintf(`LEFT OUTER JOIN "%s_labels" lt ON o.key = lt.key`, dbName)
 	}
 	params := []any{}
 
@@ -817,25 +813,6 @@ func getField(a any, field string) (any, error) {
 		}
 	}
 	return obj, nil
-}
-
-// hasNegativeExistenceTest is used for label tests, where we're looking for either
-// labels that don't exist, or aren't equal to a target value (and the interpretation
-// of a missing label is that it's *always* not equal to the target value.
-// If this is true, we need to do a `LEFT OUTER JOIN` on the main-table JOIN labels-table,
-// in order to keep all the main-table rows that don't have associated labels.
-func hasNegativeExistenceTest(orFilters []OrFilter) bool {
-	for _, orFilter := range orFilters {
-		for _, filter := range orFilter.Filters {
-			switch filter.Op {
-			case NotEq, NotExists, NotIn:
-				if isLabelFilter(&filter) {
-					return true
-				}
-			}
-		}
-	}
-	return false
 }
 
 func extractSubFields(fields string) []string {
--- a/pkg/cache/sql/informer/listoption_indexer.go
+++ b/pkg/cache/sql/informer/listoption_indexer.go
@@ -292,15 +292,20 @@ func (l *ListOptionIndexer) constructQuery(lo ListOptions, partitions []partitio
 	query += "\n  "
 	query += fmt.Sprintf(`JOIN "%s_fields" f ON o.key = f.key`, dbName)
 	if queryHasLabelFilter {
-		query += "\n  "
-		query += fmt.Sprintf(`LEFT OUTER JOIN "%s_labels" lt ON o.key = lt.key`, dbName)
+		for i, orFilters := range lo.Filters {
+			if hasLabelFilter([]OrFilter{orFilters}) {
+				query += "\n  "
+				// Make the lt index 1-based for readability
+				query += fmt.Sprintf(`LEFT OUTER JOIN "%s_labels" lt%d ON o.key = lt%d.key`, dbName, i+1, i+1)
+			}
+		}
 	}
 	params := []any{}
 
 	// 2- Filtering: WHERE clauses (from lo.Filters)
 	whereClauses := []string{}
-	for _, orFilters := range lo.Filters {
-		orClause, orParams, err := l.buildORClauseFromFilters(orFilters, dbName)
+	for i, orFilters := range lo.Filters {
+		orClause, orParams, err := l.buildORClauseFromFilters(i+1, orFilters, dbName)
 		if err != nil {
 			return queryInfo, err
 		}
@@ -537,7 +542,7 @@ func (l *ListOptionIndexer) validateColumn(column string) error {
 }
 
 // buildORClause creates an SQLite compatible query that ORs conditions built from passed filters
-func (l *ListOptionIndexer) buildORClauseFromFilters(orFilters OrFilter, dbName string) (string, []any, error) {
+func (l *ListOptionIndexer) buildORClauseFromFilters(index int, orFilters OrFilter, dbName string) (string, []any, error) {
 	var params []any
 	clauses := make([]string, 0, len(orFilters.Filters))
 	var newParams []any
@@ -546,7 +551,7 @@ func (l *ListOptionIndexer) buildORClauseFromFilters(orFilters OrFilter, dbName 
 
 	for _, filter := range orFilters.Filters {
 		if isLabelFilter(&filter) {
-			newClause, newParams, err = l.getLabelFilter(filter, dbName)
+			newClause, newParams, err = l.getLabelFilter(index, filter, dbName)
 		} else {
 			newClause, newParams, err = l.getFieldFilter(filter)
 		}
@@ -633,7 +638,7 @@ func (l *ListOptionIndexer) getFieldFilter(filter Filter) (string, []any, error)
 	return "", nil, fmt.Errorf("unrecognized operator: %s", opString)
 }
 
-func (l *ListOptionIndexer) getLabelFilter(filter Filter, dbName string) (string, []any, error) {
+func (l *ListOptionIndexer) getLabelFilter(index int, filter Filter, dbName string) (string, []any, error) {
 	opString := ""
 	escapeString := ""
 	matchFmtToUse := strictMatchFmt
@@ -647,7 +652,7 @@ func (l *ListOptionIndexer) getLabelFilter(filter Filter, dbName string) (string
 		} else {
 			opString = "="
 		}
-		clause := fmt.Sprintf(`lt.label = ? AND lt.value %s ?%s`, opString, escapeString)
+		clause := fmt.Sprintf(`lt%d.label = ? AND lt%d.value %s ?%s`, index, index, opString, escapeString)
 		return clause, []any{labelName, formatMatchTargetWithFormatter(filter.Matches[0], matchFmtToUse)}, nil
 
 	case NotEq:
@@ -662,11 +667,11 @@ func (l *ListOptionIndexer) getLabelFilter(filter Filter, dbName string) (string
 			Field: filter.Field,
 			Op:    NotExists,
 		}
-		existenceClause, subParams, err := l.getLabelFilter(subFilter, dbName)
+		existenceClause, subParams, err := l.getLabelFilter(index, subFilter, dbName)
 		if err != nil {
 			return "", nil, err
 		}
-		clause := fmt.Sprintf(`(%s) OR (lt.label = ? AND lt.value %s ?%s)`, existenceClause, opString, escapeString)
+		clause := fmt.Sprintf(`(%s) OR (lt%d.label = ? AND lt%d.value %s ?%s)`, existenceClause, index, index, opString, escapeString)
 		params := append(subParams, labelName, formatMatchTargetWithFormatter(filter.Matches[0], matchFmtToUse))
 		return clause, params, nil
 
@@ -675,18 +680,18 @@ func (l *ListOptionIndexer) getLabelFilter(filter Filter, dbName string) (string
 		if err != nil {
 			return "", nil, err
 		}
-		clause := fmt.Sprintf(`lt.label = ? AND lt.value %s ?`, sym)
+		clause := fmt.Sprintf(`lt%d.label = ? AND lt%d.value %s ?`, index, index, sym)
 		return clause, []any{labelName, target}, nil
 
 	case Exists:
-		clause := fmt.Sprintf(`lt.label = ?`)
+		clause := fmt.Sprintf(`lt%d.label = ?`, index)
 		return clause, []any{labelName}, nil
 
 	case NotExists:
 		clause := fmt.Sprintf(`o.key NOT IN (SELECT o1.key FROM "%s" o1
 		JOIN "%s_fields" f1 ON o1.key = f1.key
-		LEFT OUTER JOIN "%s_labels" lt1 ON o1.key = lt1.key
-		WHERE lt1.label = ?)`, dbName, dbName, dbName)
+		LEFT OUTER JOIN "%s_labels" lt%di1 ON o1.key = lt%di1.key
+		WHERE lt%di1.label = ?)`, dbName, dbName, dbName, index, index, index)
 		return clause, []any{labelName}, nil
 
 	case In:
@@ -695,7 +700,7 @@ func (l *ListOptionIndexer) getLabelFilter(filter Filter, dbName string) (string
 			target += strings.Repeat(", ?", len(filter.Matches)-1)
 		}
 		target += ")"
-		clause := fmt.Sprintf(`lt.label = ? AND lt.value IN %s`, target)
+		clause := fmt.Sprintf(`lt%d.label = ? AND lt%d.value IN %s`, index, index, target)
 		matches := make([]any, len(filter.Matches)+1)
 		matches[0] = labelName
 		for i, match := range filter.Matches {
@@ -713,11 +718,11 @@ func (l *ListOptionIndexer) getLabelFilter(filter Filter, dbName string) (string
 			Field: filter.Field,
 			Op:    NotExists,
 		}
-		existenceClause, subParams, err := l.getLabelFilter(subFilter, dbName)
+		existenceClause, subParams, err := l.getLabelFilter(index, subFilter, dbName)
 		if err != nil {
 			return "", nil, err
 		}
-		clause := fmt.Sprintf(`(%s) OR (lt.label = ? AND lt.value NOT IN %s)`, existenceClause, target)
+		clause := fmt.Sprintf(`(%s) OR (lt%d.label = ? AND lt%d.value NOT IN %s)`, existenceClause, index, index, target)
 		matches := append(subParams, labelName)
 		for _, match := range filter.Matches {
 			matches = append(matches, match)
--- a/pkg/cache/sql/informer/listoption_indexer.go
+++ b/pkg/cache/sql/informer/listoption_indexer.go
@@ -100,7 +100,8 @@ func NewListOptionIndexer(fields [][]string, s Store, namespaced bool) (*ListOpt
 		namespaced:    namespaced,
 		indexedFields: indexedFields,
 	}
-	l.RegisterAfterUpsert(l.afterUpsert)
+	l.RegisterAfterUpsert(l.addIndexFields)
+	l.RegisterAfterUpsert(l.addLabels)
 	l.RegisterAfterDelete(l.afterDelete)
 	columnDefs := make([]string, len(indexedFields))
 	for index, field := range indexedFields {
@@ -179,8 +180,8 @@ func NewListOptionIndexer(fields [][]string, s Store, namespaced bool) (*ListOpt
 
 /* Core methods */
 
-// afterUpsert saves sortable/filterable fields into tables
-func (l *ListOptionIndexer) afterUpsert(key string, obj any, tx db.TXClient) error {
+// addIndexFields saves sortable/filterable fields into tables
+func (l *ListOptionIndexer) addIndexFields(key string, obj any, tx db.TXClient) error {
 	args := []any{key}
 	for _, field := range l.indexedFields {
 		value, err := getField(obj, field)
@@ -213,14 +214,11 @@ func (l *ListOptionIndexer) afterUpsert(key string, obj any, tx db.TXClient) err
 	if err != nil {
 		return &db.QueryError{QueryString: l.addFieldQuery, Err: err}
 	}
-	return l.upsertLabels(tx, key, obj)
+	return nil
 }
 
-func (l *ListOptionIndexer) upsertLabels(tx db.TXClient, key string, obj any) error {
-	return l.UpsertLabels(tx, l.upsertLabelsStmt, key, obj, l.GetShouldEncrypt())
-}
-
-func (l *ListOptionIndexer) UpsertLabels(tx db.TXClient, stmt *sql.Stmt, key string, obj any, shouldEncrypt bool) error {
+// labels are stored in tables that shadow the underlying object table for each GVK
+func (l *ListOptionIndexer) addLabels(key string, obj any, tx db.TXClient) error {
 	k8sObj, ok := obj.(*unstructured.Unstructured)
 	if !ok {
 		logrus.Debugf("UpsertLabels: Error?: Can't convert obj into an unstructured thing.")
@@ -228,7 +226,7 @@ func (l *ListOptionIndexer) UpsertLabels(tx db.TXClient, stmt *sql.Stmt, key str
 	}
 	incomingLabels := k8sObj.GetLabels()
 	for k, v := range incomingLabels {
-		err := tx.StmtExec(tx.Stmt(stmt), key, k, v)
+		err := tx.StmtExec(tx.Stmt(l.upsertLabelsStmt), key, k, v)
 		if err != nil {
 			return err
 		}
--- a/pkg/cache/sql/informer/listoption_indexer.go
+++ b/pkg/cache/sql/informer/listoption_indexer.go
@@ -665,10 +665,18 @@ func (l *ListOptionIndexer) getLabelFilter(filter Filter) (string, []any, error)
 		clause := fmt.Sprintf(`lt.label = ?`)
 		return clause, []any{labelName}, nil
 
+	case NotExists:
+		clause := fmt.Sprintf(`lt.label != ?`)
+		return clause, []any{labelName}, nil
+
 	case In:
 		fallthrough
 	case NotIn:
-		target := fmt.Sprintf("(?%s)", strings.Repeat(", ?", len(filter.Matches)))
+		target := "(?"
+		if len(filter.Matches) > 0 {
+			target += strings.Repeat(", ?", len(filter.Matches)-1)
+		}
+		target += ")"
 		opString = "IN"
 		if filter.Op == NotIn {
 			opString = "NOT IN"
--- a/pkg/cache/sql/informer/listoption_indexer.go
+++ b/pkg/cache/sql/informer/listoption_indexer.go
@@ -42,6 +42,12 @@ var (
 )
 
 const (
+	objectDBAlias = "o"
+	fieldDBAlias  = "f"
+	labelsDBAlias = "lt" // because "l" can look too much like the numeral "1"
+)
+
+const (
 	matchFmt             = `%%%s%%`
 	strictMatchFmt       = `%s`
 	createFieldsTableFmt = `CREATE TABLE "%s_fields" (
@@ -198,10 +204,19 @@ func (l *ListOptionIndexer) afterDelete(key string, tx db.TXClient) error {
 //   - a continue token, if there are more pages after the returned one
 //   - an error instead of all of the above if anything went wrong
 func (l *ListOptionIndexer) ListByOptions(ctx context.Context, lo ListOptions, partitions []partition.Partition, namespace string) (*unstructured.UnstructuredList, int, string, error) {
-	// 1- Intro: SELECT and JOIN clauses
-	query := fmt.Sprintf(`SELECT o.object, o.objectnonce, o.dekid FROM "%s" o`, db.Sanitize(l.GetName()))
-	query += "\n  "
-	query += fmt.Sprintf(`JOIN "%s_fields" f ON o.key = f.key`, db.Sanitize(l.GetName()))
+	// 1- First, what kind of filtering will be doing?
+	hasFieldFilter, hasLabelFilter := categorizeFilters(lo.Filters)
+	dbName := db.Sanitize(l.GetName())
+	// 1.1- Intro: SELECT and JOIN clauses
+	query := fmt.Sprintf(`SELECT %s.object, %s.objectnonce, %s.dekid FROM "%s" %s`, objectDBAlias, objectDBAlias, objectDBAlias, dbName, objectDBAlias)
+	if hasFieldFilter || !hasLabelFilter {
+		query += "\n  "
+		query += fmt.Sprintf(`JOIN "%s_fields" %s ON %s.key = %s.key`, dbName, fieldDBAlias, objectDBAlias, fieldDBAlias)
+	}
+	if hasLabelFilter {
+		query += "\n  "
+		query += fmt.Sprintf(`JOIN "%s_labels" %s ON %s.key = %s.key`, dbName, labelsDBAlias, objectDBAlias, labelsDBAlias)
+	}
 	params := []any{}
 
 	// 2- Filtering: WHERE clauses (from lo.Filters)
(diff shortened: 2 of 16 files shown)


pkg/cache/sql/store/store.go
```diff
@@ -27,6 +27,15 @@ const (
...
 	)`
+	// 'key' contains values from the column key of table nodes
+	createLabelsTableFmt = `CREATE TABLE IF NOT EXISTS "%s_labels" (
+		key TEXT NOT NULL REFERENCES "%s"(key) ON DELETE CASCADE,
+		label TEXT NOT NULL,
+		value TEXT NOT NULL,
+		PRIMARY KEY (key, label)
+	)`
+	upsertLabelsStmtFmt = `REPLACE INTO "%s_labels"(key, label, value) VALUES (?, ?, ?)`
+	deleteLabelsStmtFmt = `DELETE FROM "%s_labels" WHERE KEY = ?`
```
moio, line 38: sql/store/store was designed to be a drop-in replacement to client-go's store type, only backed by SQLite instead of in-memory structures.

It sounds a bit out of place to me to bake labels in here.

Instead, I'd recommend to add it either to ListOptionIndexer or, potentially, a brand new Storage class specifically for labels (altohugh, in that case, logic would still bleed into ListOptionIndexer).

Please take a look at https://github.com/rancher/lasso/blob/master/pkg/cache/sql/informer/indexer.go for inspiration. There Indexer builds upon Store by creating own tables, inserting data into them via RegisterAfterUpsert and then querying in own methods. I think that is similar to what you are achieving here.

pkg/cache/sql/db/client.go
```diff
@@ -274,6 +276,22 @@ func (c *Client) Upsert(tx TXClient, stmt *sql.Stmt, key string, obj any, should
...
+	}
+	incomingLabels := k8sObj.GetLabels()
+	for k, v := range incomingLabels {
+		err := tx.StmtExec(tx.Stmt(stmt), key, k, v)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
moio, line 293: db/client was designed to be a wrapper around relatively low-level database operations, and should in my eyes be kept independent from logic which depends on the exact table structure ("application concerns").

I think this code belongs to the same place where the bulk of the logic for labels goes, see the discussion about store below.

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -41,6 +41,12 @@ var (
 	InvalidColumnErr = errors.New("supplied column is invalid")
 )
 
+const (
+	objectDBAlias = "o"
+	fieldDBAlias  = "f"
+	labelsDBAlias = "lt" // because "l" can look too much like the numeral "1"
```
moio, line 47: I'd suggest baking those aliases directly in SQL code, interpolation makes it a little bit more difficult to understand what's going on in my eyes

moio (changes requested):
I left some inline comments, overall this seems to be going in a good direction, thanks for the contribution.

moio:
One other thing: can you please open the corresponding Steve PR as well, even in draft mode?

I'd like to have a look at the whole picture, and possibly locally try it out as well.

Thanks in advance!

pkg/cache/sql/informer/listoption_indexer.go
ericpromislow: ok.  I did that because I'm never thrilled to see an "o", "O", or "l" in code by itself, but I'll revert that.

pkg/cache/sql/store/store.go
ericpromislow: k, I moved it into the listoption indexer and it seems to fit right.

ericpromislow:
The steve PR is at https://github.com/rancher/steve/pull/317

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -51,8 +55,21 @@ const (
 	createFieldsIndexFmt = `CREATE INDEX "%s_%s_index" ON "%s_fields"("%s")`
 
 	failedToGetFromSliceFmt = "[listoption indexer] failed to get subfield [%s] from slice items: %w"
+
+	// 'key' contains values from the column key of table nodes
```
moio, line 59: Nitpick: why "nodes"?

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -173,6 +201,26 @@ func (l *ListOptionIndexer) afterUpsert(key string, obj any, tx db.TXClient) err
...
 		return &db.QueryError{QueryString: l.addFieldQuery, Err: err}
 	}
+	return l.upsertLabels(tx, key, obj)
+}
+
+func (l *ListOptionIndexer) upsertLabels(tx db.TXClient, key string, obj any) error {
+	return l.UpsertLabels(tx, l.upsertLabelsStmt, key, obj, l.GetShouldEncrypt())
+}
+
+func (l *ListOptionIndexer) UpsertLabels(tx db.TXClient, stmt *sql.Stmt, key string, obj any, shouldEncrypt bool) error {
```
moio, line 211: Two observations:

1. it seems like upsertLabels is only really called here, and does not add much to UpsertLabels. I'd suggest to unify the two methods into one
2. I would suggest to uniform how afterUpsert is structured, by having it call two methods: one that encapsulates current functionality (basically `INSERT INTO "%s_fields"`) and another one for the new bits (`upsertLabels`)

That way it should be easier to read and possible to test in isolation as well.


Same for `afterDelete` below.

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -51,8 +55,21 @@ const (
...
 
 	failedToGetFromSliceFmt = "[listoption indexer] failed to get subfield [%s] from slice items: %w"
+
+	// 'key' contains values from the column key of table nodes
+	createLabelsTableFmt = `CREATE TABLE IF NOT EXISTS "%s_labels" (
+		key TEXT NOT NULL REFERENCES "%s"(key) ON DELETE CASCADE,
+		label TEXT NOT NULL,
+		value TEXT NOT NULL,
+		PRIMARY KEY (key, label)
+	)`
```
moio, line 65: We will need at least an `INDEX` over `label` here, to speed up queries at retrieval time.

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -405,37 +467,69 @@ func (l *ListOptionIndexer) validateColumn(column string) error {
...
+
+	clause := fmt.Sprintf(`f."%s" %s ? ESCAPE '\'`, columnName, opString)
+	return clause, []any{formatMatchTarget(filter)}, nil
+}
+
+func (l *ListOptionIndexer) getLabelFilter(filter Filter) (string, []any, error) {
+	opString := "LIKE"
+	if filter.Op == NotEq {
+		opString = "NOT LIKE"
+	}
```
moio, line 509: I fear this will not suffice.

Per the "Proposed solution" paragraph in [the issue](https://github.com/rancher/rancher/issues/46333#issue-2429999150) we'd really need the full label selector syntax and power, per https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors - with some kind of parser on the Steve side.

moio (changes requested):
Thanks, this gets closer. I found two issues, one of which (adding an INDEX) should be very straightforward to solve.

The other will likely need changes in the parser in Steve as well as changes in the query builder here in listOptionsIndexer.

Thanks, keep up the good work!

pkg/cache/sql/informer/listoption_indexer.go
ericpromislow: "Nodes" is cruft from earlier code, and I don't remember why I picked it.  I'll deal with that next tomorrow (Friday Nov 8)

pkg/cache/sql/informer/listoption_indexer.go
ericpromislow: I added that today.

pkg/cache/sql/informer/listoption_indexer.go
ericpromislow: Friday work…

pkg/cache/sql/informer/listoption_indexer.go
ericpromislow: Thanks ... I missed that note. Yes, I'll add a proper op parser.

pkg/cache/sql/informer/listoption_indexer.go
ericpromislow: There's no need for a comment at all -- it was restating the DDL, and went out of date.

pkg/cache/sql/informer/listoption_indexer.go
ericpromislow: Yes. And `afterUpsert` and `afterDelete` are bad names for the actual methods.

pkg/cache/sql/informer/listoption_indexer.go
ericpromislow: Do we really need to implement the entire Label Selector command set from kubectl? Because here isn't the only client the UI, and which operators does it need?

^ @richard-cox

ericpromislow: It's easy enough to call `(k8s.io/apimachinery/pkg/labels/selector.go) labels.Parse()` or `labels.ParseToRequirements()` to pass in a standard k8s label selector expression and get back a list of requirements, but I'm not convinced it's needed.

richard-cox: Unfortunately full label selector support is needed (second point in https://github.com/rancher/rancher/issues/46333#issuecomment-2250742787). The UI has a really useful feature used in a number of places to show the result of label selector config. At the moment it is done locally on all resources, which is a scaling nightmare. It needs to change to on-demand via the new API.

moio: @ericpromislow good that we can have the parsing done easily

ericpromislow: I need to see if there are any conflicts between the current filter language and the kubectl label selection language.

ericpromislow: @moio @richard-cox 

So there are several problems with using the kubernetes label parser.

First, it requires values to match (the right hand part of '=' or another match operator) the
k8s label syntax as described at https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set 

Second, quotations aren't allowed in the target values, so we can't rely on the label parser to
determine whether to do an exact match or not.

It might make sense to do label parsing only for these sub-queries:

`metadata.labels.NAME <REST-OF-SUBQUERY>`
and
`metadata.labels.[IDENTIFIER] <REST-OF-SUBQUERY>`

and then we have to preprocess the expression a bit into the format
`<NAME | IDENTIFIER> <REST-OF-SUBQUERY>`

Now if you want to support all the k8s label-parser operators:

```
=
!=
EXISTS
! EXISTS
IN (set)
NOTIN (set)
```

we need to distinguish commas inside sets from commas that act like an or-operator.
I've pushed a failing test to https://github.com/ericpromislow/steve/blob/50dbbdd4000e31051123856729aee7075d19d480/pkg/stores/sqlpartition/listprocessor/processor_test.go#L269-L295

We also need to add quotation marks as delimiters to indicate exact match.

The ad-hoc parser doesn't handle the existence and set-membership tests, so I'm going to copy the
k8s parser (Apache-licensed) into our code and modify it. I'm not worried about forking it -- the label
selection syntax is unlikely to change in the near term, and if it does it shouldn't be difficult to incorporate the changes.

ericpromislow:
This change has the new filter query expression parser, with tests.  There's some cruft that came in from the k8s code that I'll pull out during the next round of reviews.

pkg/cache/sql/informer/listoption_indexer.go
richard-cox: Would you be able to provide some sample url params demonstrating how the label selector would work?

UI side i don't think we're tied to a specific format, we would need to convert the json representation of the label selector to params strings in all worlds. Does that help at all?

moio: @ericpromislow how about making `labelSelector` a standalone filter query parameter, on par with `filter`, `sort`, `limit`, etc?

If we do so, commas retain their original Steve meaning in the scope of `filter` (`OR` between filters) and the Kubernetes standard meaning in the scope of `labelSelector` (`AND` between selectors or set-element separator - it'll be up to the labelSelector parsing code to figure that out). It will be implied that any `labelSelector` acts on `metadata.labels` - it's implicit, like everywhere else in Kubernetes.

Rewording my question: could we add a top-level `labelSelector` parameter and then treat its value as a black box passing it straight to Kubernetes parsing code and not worry about anything else?

moio: References:
 - https://github.com/rancher/steve?tab=readme-ov-file#list-specific-query-parameters
 - https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors

richard-cox: FWIW the UI will be taking selectors config from https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#labelselector-v1-meta and making a request to get the result

labelSelectors contain two parts
- array of matchExpression. basically `key` - `operator` (in, not in, less than, etc), - `values`
- matchLabel is shorthand for `key` - `IN` - `values`

ericpromislow: There's only so much of a reason to use the k8s label selectors as is, and that's because we aren't using the k8s matching mechanism (also in apimachinery), but converting the parsed queries into SQL queries. I'm doing that now with the forked parser.

The only mod I needed to make to the label query parser was support for single-quoted and double-quoted strings. And we get errors on unterminated strings basically for free (including `...\\` at the end of input).

I believe I've correctly implemented everything for https://github.com/rancher/steve?tab=readme-ov-file#filter

I haven't implemented support for less-than and greater-than yet (the k8s language predictably uses "<" and ">"). Right now they're parsed and the code that converts a k8s `Requirement` into an `informer.Filter` throws an error when it sees one of those. Let me go build a downstream cluster so I can see what happens with selecting `provisioning.cattle.io/v1/Cluster`s

moio: > Note that for non-identifier target values, they need to be quoted, so you can't do a LIKE match on a target value like we contain spaces. You can do a LIKE match on one of its words, or a subword, but not anything more complex than a single identifier-type token.

@richard-cox is this acceptable from your perspective?

In particular, where does the UI get those label selectors from? Are they from user input, are they constructed by UI code from user input, or do they come from other Kubernetes objects that have labelSelectors in them, such as these https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements?

IOW how close of a syntax match do you need. Does it really have to match 1:1 100% with the Kubernetes spec?

richard-cox: > In particular, where does the UI get those label selectors from? Are they from user input, are they constructed by UI code from user input, or do they come from other Kubernetes objects that have labelSelectors in them

All of the above. Use cases are listed in https://github.com/rancher/dashboard/issues/10417#issue-2123233258 but very generally

1. User is creating a new labelSelector object via input fields and the UI will show the result of those proposed selectors (it will then persist the changes as a labelSelector object within the resource)
2. Applying the labelSelector from within a resource to find other resources

So the UI isn't using label selector like functionality, it's literally label selector stuff. Being able to match the exact functionality is important. It would be a collection of [these](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#labelselectorrequirement-v1-meta) AND'd together

key: string
operator: In, NotIn, Exists and DoesNotExist
values: string array

> Note there is also [Node labelSelector](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#labelselectorrequirement-v1-meta) requirements that contain Gt (greater than) and Lt (less than)

Examples
```
foo IN [bar]
foo IN [bar, baz2]
aaa NotIn [bar, baz2]
bbb Exists
ccc DoesNotExist
ddd Gt 1
eee Lt 2
```

which i think equates to 

```
  labelSelector
    matchLabels:
     foo: bar
    matchExpressions:
     - key: foo
       op: In
       values: [bar, baz]
     - key: aaa
       op: NotIn
       values: [bar, baz]
     - key: bbb
       op: Exists
     - key: ccc
       op: DoesNotExist
     - key: ddd
       op: Gt
       values: 1
     - key: eee
       op: Lt
       values: 2
```

@ericpromislow thanks for providing an example below (`filter=metadata.labels%5bauthz.management.cattle.io/default-project%5d="true"'`) can you provide filter param equivalents for the examples above?

> can't do a LIKE match on a target value like we contain spaces. You can do a LIKE match on one of its words, or a subword, but not anything more complex than a single identifier-type token.

I couldn't find LIKE functionality in labelSelector land, so as long as it doesn't break existing functionality i think that's fine

ericpromislow: @richard-cox  Here how you would encode those above examples:
```
foo IN [bar] => ?filter=foo+IN+(bar)
foo IN [bar, baz2] => ?filter=foo+IN+(bar,baz2) 
aaa NotIn [bar, baz2]=> ?filter=foo+NOTIN+(bar,baz2) 
bbb Exists=> ?filter=bbb
ccc DoesNotExist ?filter=!bbb. # or %21bbb
```

General comments:
1. The case of the keywords is always ignore, so something like `eXiSt` is ok, but I use ALLCAPS for clarity.
2. White space is ignored outside strings.  The second request could equivalently be
`?filter=foo+IN+++(+bar+,++++++baz2++++)`

3. The `EXISTS` and `NOT EXISTS` implementations don't make a lot of sense in terms of what gets returned. Right now `EXISTS` returns the row if it's matched, and in fact I can't get `NOT EXISTS` to work.

But I think it would make more sense if the two existence tests just return something simpler than a row-set ... I'll need to work on that.

The next two haven't been implemented yet, but should be:
```
ddd Gt 1 => ?filter=ddd+<+1
eee Lt 2=> ?filter=eee+>+2
```

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -427,38 +534,179 @@ func (l *ListOptionIndexer) validateColumn(column string) error {
...
 
-		orWhereClause += fmt.Sprintf(`f."%s" %s ? ESCAPE '\'`, columnName, opString)
-		format := strictMatchFmt
+//type Filter struct {
+//	Field   []string
+//	Match   []string
+//	Op      Op: one of
+//	Partial bool
+//}
+
```
moio, line 565:
```suggestion
```

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -427,38 +534,179 @@ func (l *ListOptionIndexer) validateColumn(column string) error {
...
+		} else {
+			opString = "!="
+		}
+		clause := fmt.Sprintf(`f."%s" %s ?%s`, columnName, opString, escapeString)
+		return clause, []any{formatMatchTarget(filter)}, nil
+	case Exists:
+		clause := fmt.Sprintf(`f."%s" IS NOT NULL`, columnName)
+		return clause, []any{}, nil
+	case NotExists:
+		clause := fmt.Sprintf(`f."%s" IS NULL`, columnName)
```
moio, line 604: I fear this will never happen, because addIndexedField will use empty string instead of proper `NULL`:

https://github.com/rancher/lasso/pull/117/files?w=1#diff-910681530ceac7cfe911fb307754542734adee41c8d009d795aabe0d9e3803b2R200

To do this right we have to:
 - really insert `NULL`
 - check for nullability every time we do another comparision (eg. `=` or `<>`, because [3-valued logic](https://en.wikipedia.org/wiki/Null_(SQL)#Comparisons_with_NULL_and_the_three-valued_logic_(3VL)))
 - make sure this does not break the UI

Before implementing that, though, do we need new operator support in the UI at all?

IIUC:
 - Before all we had for fields was equality and inequality
 - This PR's main intent is about adding support for filtering by labels and with more operators (in, not in, exists, doesn't exist)
 - This code section basically adds the same operators for fields as well

@ericpromislow is that correct?
@richard-cox do we need them?

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -427,38 +534,179 @@ func (l *ListOptionIndexer) validateColumn(column string) error {
...
+
+	return "", nil, fmt.Errorf("unrecognized operator: %s", opString)
+}
+
+func (l *ListOptionIndexer) getLabelFilter(filter Filter, dbName string) (string, []any, error) {
+	opString := ""
+	escapeString := ""
+	if len(filter.Field) < 3 || filter.Field[0] != "metadata" || filter.Field[1] != "labels" {
+		return "", nil, fmt.Errorf("expecting a metadata.labels field, got '%s'", strings.Join(filter.Field, "."))
+	}
```
moio, line 633:
```suggestion
```

I feel like this is a bit of defensive programming: `getLabelFilter` is unexporter and only called by `buildORClauseFromFilters`, which uses `isLabelFilter` immediately before the call to check this very condition (well, its negation).

Am I missing anything?

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -427,38 +534,179 @@ func (l *ListOptionIndexer) validateColumn(column string) error {
...
+	matchFmtToUse := strictMatchFmt
+	labelName := filter.Field[2]
+	switch filter.Op {
+	case Eq:
+		if filter.Partial {
+			opString = "LIKE"
+			escapeString = escapeBackslashDirective
```
moio, line 640: How about inlining the directive in the SQL query where it's used, instead of having this as a constant copied into a variable and then into the query?

(yes, I understand that's a bit of repetition, maybe it's just me suffering from indirection motion sickness!)

PS. same about matchFmt. I had a hard time looking into 4 places before coalescing the sense of the query in my mind while reviewing 🦀

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -186,25 +241,57 @@ func (l *ListOptionIndexer) afterDelete(key string, tx db.TXClient) error {
...
+
+	// 1- First, what kind of filtering will be doing?
+	// 1.1- Intro: SELECT and JOIN clauses
+	query := fmt.Sprintf(`SELECT o.object, o.objectnonce, o.dekid FROM "%s" o`, dbName)
 	query += "\n  "
-	query += fmt.Sprintf(`JOIN "%s_fields" f ON o.key = f.key`, db.Sanitize(l.GetName()))
+	query += fmt.Sprintf(`JOIN "%s_fields" f ON o.key = f.key`, dbName)
+	if hasLabelFilter(lo.Filters) {
+		query += "\n  "
+		query += fmt.Sprintf(`JOIN "%s_labels" lt ON o.key = lt.key`, dbName)
```
moio, line 285: I fear this will not always work.

Main failure case will be when a single object row has multiple labels: `JOIN` will create multiple resulting rows with all object/field columns repeated, and different label columns.

Secondarily, if an object has no labels at all then it will be filtered out by this `JOIN`.


One way to resolve this problem could be to not JOIN the labels table at all here, and only use subqueries in filters when needed.

Another is to `LEFT OUTER JOIN` here, and then deduplicate by DISTINCT on all other columns before returning results. Probably more convoluted.


Do we have any tests involving the database that cover this?

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -176,7 +214,24 @@ func (l *ListOptionIndexer) afterUpsert(key string, obj any, tx db.TXClient) err
...
+	k8sObj, ok := obj.(*unstructured.Unstructured)
+	if !ok {
+		logrus.Debugf("UpsertLabels: Error?: Can't convert obj into an unstructured thing.")
+		return nil
+	}
+	incomingLabels := k8sObj.GetLabels()
+	for k, v := range incomingLabels {
+		err := tx.StmtExec(tx.Stmt(l.upsertLabelsStmt), key, k, v)
+		if err != nil {
+			return err
```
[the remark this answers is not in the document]
ericpromislow: Straightforward to do, but I don't see a straightforward way to do unit tests with that change because it's run from an `afterUpsert` handler in a different module.

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -176,7 +214,24 @@ func (l *ListOptionIndexer) afterUpsert(key string, obj any, tx db.TXClient) err
 	return nil
 }
 
-func (l *ListOptionIndexer) afterDelete(key string, tx db.TXClient) error {
+// labels are stored in tables that shadow the underlying object table for each GVK
+func (l *ListOptionIndexer) addLabels(key string, obj any, tx db.TXClient) error {
+	k8sObj, ok := obj.(*unstructured.Unstructured)
+	if !ok {
+		logrus.Debugf("UpsertLabels: Error?: Can't convert obj into an unstructured thing.")
```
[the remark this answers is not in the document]
ericpromislow: FWICT this is what happens.  In store.go:upsert():
```
tx := s.BeginTx(...)
err = s.Upsert()...
return on err
err = s.runAfterUpsert()
return on err
tx.Commit()
```

So yes, right now if `afterUpsert` fails, the initial `upsert` is rolled back as well.

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -186,25 +241,57 @@ func (l *ListOptionIndexer) afterDelete(key string, tx db.TXClient) error {
...
 
+func (l *ListOptionIndexer) deleteLabels(key string, tx db.TXClient) error {
+	err := tx.StmtExec(tx.Stmt(l.deleteLabelsStmt), key)
+	if err != nil {
+		return &db.QueryError{QueryString: l.deleteLabelsQuery, Err: err}
+	}
+	return nil
+}
+
+type QueryInfo struct {
```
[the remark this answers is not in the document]
ericpromislow: done

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -352,14 +439,30 @@ func (l *ListOptionIndexer) ListByOptions(ctx context.Context, lo ListOptions, p
 		offsetClause = "\n  OFFSET ?"
 		params = append(params, offset)
 	}
+	if limit == 0 && offset == 0 {
+		countQuery = ""
+		countParams = []any{}
+	}
```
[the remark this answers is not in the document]
ericpromislow: In the current implementation, if `limit == 0 && offset == 0` we don't run a count statement.

Also in the current implementation, constructing the SQL query and running it are done in one function,
so they're interleaved, which makes query generation more cumbersome to test -- you need to test it
through mocks, rather than just looking at the string a function call returns.

I wanted to move all the query construction into one function, so the best way to tell the executor not to
run the count query is to set it to the empty string.

So I maintain the new code isn't changing the database lookups.

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -186,25 +241,57 @@ func (l *ListOptionIndexer) afterDelete(key string, tx db.TXClient) error {
...
+		return nil, 0, "", err
+	}
+	return l.executeQuery(ctx, queryInfo)
+}
+
+func (l *ListOptionIndexer) constructQuery(lo ListOptions, partitions []partition.Partition, namespace string, dbName string) (*QueryInfo, error) {
+	queryInfo := &QueryInfo{}
+
+	// 1- First, what kind of filtering will be doing?
+	// 1.1- Intro: SELECT and JOIN clauses
```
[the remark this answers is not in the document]
ericpromislow: Done

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -352,14 +439,30 @@ func (l *ListOptionIndexer) ListByOptions(ctx context.Context, lo ListOptions, p
...
+		countParams: countParams,
+		limit:       limit,
+		offset:      offset,
+	}
+	return queryInfo, nil
+}
 
+func (l *ListOptionIndexer) executeQuery(ctx context.Context, queryInfo *QueryInfo) (*unstructured.UnstructuredList, int, string, error) {
 	// execute
+	query := queryInfo.query
```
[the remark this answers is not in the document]
ericpromislow: Yup

pkg/cache/sql/informer/listoption_indexer.go
ericpromislow: Right. I was using that as a guide while writing the code. Not needed now.

pkg/cache/sql/informer/listoption_indexer.go
ericpromislow: Now that I think about it, my guess is that NULL/!NULL don't make sense for the non-labels, because either we index them (which means they can't be NULL, because they always come with a value, right?) or we don't index them, which means you can't query on those fields.

But NULL/!NULL do make sense for labels, because you're basically querying on the existence of a key in a hash. Also it took me a couple of hours to work out the syntax for `NotExists(label)` so I don't want to toss it that quickly.

pkg/cache/sql/informer/listoption_indexer.go
ericpromislow: Yes, exactly. I believe I wrote the code in `getLabelFilter` first and then filled in the other details. So this was scaffolding that can be deleted.

pkg/cache/sql/informer/listoption_indexer.go
ericpromislow: # TBD
We need to hear from Richard if this business with quoting strings to be exact vs doing substring matches is going to survive.

pkg/cache/sql/informer/listoption_indexer.go
ericpromislow: My main peeve here is that we're mocking the database in the unit tests, and that makes increasing less sense to me.

I've written plenty of unit tests in Rails and Flask (python), and you never mock the database.  You just use a test variant of it, and the framework resets the database on every test. Yes, it takes a bit longer, but the tests are much simpler (no mocks) and you have more confidence that the DB code will run correctly in production.

So, no, we don't have any tests that cover this. 

Yes, I'm proposing that we rewrite the unit tests .

In my defense: https://markphelps.me/posts/writing-tests-for-your-database-code-in-go/ : 
```
However, I would argue that mocking the database when testing your SQL code is an anti-pattern that you should avoid, mainly because it violates requirement 3[*] since it isn’t actually testing that we can correctly interact with a real database as we are mocking the database itself!
```
* "We want to test that we can correctly interact with the database and also test that the business logic is correct."

I certainly worked out my statements interactively with a sqlite DB, namely the one created by running steve while creating a few events, and then shutting down steve so the DB wouldn't change.

Here's an example:
```
sqlite> select * from _v1_Event_labels;
default/my-custom-event4|app|app1
default/my-custom-event4|bar|chocolate
default/my-custom-event5|app|app1
default/my-custom-event5|bar|chocolate
default/my-custom-event5|cat|morris
default/my-custom-event5|dog|zipher
sqlite> SELECT o.object, o.objectnonce, o.dekid FROM "_v1_Event" o JOIN "_v1_Event_fields" f on o.key = f.key JOIN "_v1_Event_labels" lt on o.key = lt.key WHERE (lt.label = "dog" and lt.value = "zipher");
%
 Unstructured??||0
```

This looks ok. Unfortunately the fields we're asking for our partly binary and not very useful.  But let's look at a label that is used more than once:
```
sqlite> SELECT o.object, o.objectnonce, o.dekid FROM "_v1_Event" o JOIN "_v1_Event_fields" f on o.key = f.key JOIN "_v1_Event_labels" lt on o.key = lt.key WHERE (lt.label = "bar" and lt.value = "chocolate");
%
 Unstructured??||0
%
 Unstructured??||0
```

But it's hard to tell what we're getting, so let's get the key instead:

```
sqlite> SELECT o.key FROM "_v1_Event" o JOIN "_v1_Event_fields" f on o.key = f.key JOIN "_v1_Event_labels" lt on o.key = lt.key WHERE (lt.label = "bar" and lt.value = "chocolate");
default/my-custom-event4
default/my-custom-event5
```

Unless I'm missing something, this looks ok.  In the next comment I'll report on an object where there are no labels.

ericpromislow: So here's a `reason="Starting"` event for `metadata.name=lima-rancher-desktop.17e053aa93e89c71` that has no labels.  Here are two curl calls, one without labels, one with, on this event:
```
$ curl -ksL https://localhost:5111/v1/events'?filter=metadata.name="lima-rancher-desktop.17e053aa93e89c71"' | jq . | wc
      94     165    2826
$ curl -ksL https://localhost:5111/v1/events'?filter=metadata.name="lima-rancher-desktop.17e053aa93e89c71"&filter=metadata.labels.fish=car' | jq .
{
  "type": "collection",
  "links": {
    "self": "https://localhost:5111/v1/events"
  },
  "createTypes": {
    "event": "https://localhost:5111/v1/events"
  },
  "actions": {},
  "resourceType": "event",
  "data": []
}
```

The SQL for the second call was this:
```
SELECT o.object, o.objectnonce, o.dekid FROM "_v1_Event" o
  JOIN "_v1_Event_fields" f ON o.key = f.key
  JOIN "_v1_Event_labels" lt ON o.key = lt.key
  WHERE
    (f."metadata.name" = "lima-rancher-desktop.17e053aa93e89c71") AND
    (lt.label = "fish" AND lt.value LIKE "%car% ESCAPE '\')
  ORDER BY f."metadata.namespace" ASC, f."metadata.name" ASC 
  LIMIT 100000
``` 

and it behaved as expected.

pkg/cache/sql/informer/listoption_indexer.go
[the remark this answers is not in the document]
moio: I do not see how changing the returned error type should make testing any harder - if anything, in my mind it should make it easier to detect.

What am I missing?

pkg/cache/sql/informer/listoption_indexer.go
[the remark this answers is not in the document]
moio: Sorry, I fail to understand.

If the cast at L224 fails then nil is returned by L227 so the caller, `runAfterUpsert`, will also return `nil` and therefore `upsert` will commit (assuming no errors from other registered after-upsert functions).

What I believe should happen here is that the entire transaction (upsert + any registered after-upsert functions) should be rolled back.

What am I missing?

pkg/cache/sql/informer/listoption_indexer.go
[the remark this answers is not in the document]
moio: > In the current implementation, if limit == 0 && offset == 0 we don't run a count statement.

Agreed

> Also in the current implementation, constructing the SQL query and running it are done in one function,
so they're interleaved, which makes query generation more cumbersome to test -- you need to test it
through mocks, rather than just looking at the string a function call returns.

Agreed

> I wanted to move all the query construction into one function

Agreed, it's cleaner now

> So I maintain the new code isn't changing the database lookups.

Agreed

> However, I noticed that there's no need to include the ORDER clause in a SELECT COUNT stmt, so I dropped it.

Agreed

> the best way to tell the executor not to run the count query is to set it to the empty string

Meh... I personally prefer not to mess with mutation unless necessary, so I'd rather have all struct members be written once. Also, an argument could be made that it's the job of the executor to execute - and so potentially to decide when not to.

Anyway, I understand taste and style play a role here. Up to you to make the final call to either change this or just resolve this conversation.

pkg/cache/sql/informer/listoption_indexer.go
moio: ACK

pkg/cache/sql/informer/listoption_indexer.go
moio: > My main peeve here is that we're mocking the database in the unit tests, and that makes increasing less sense to me.

100% agreed. DB should be in the loop - we do care about correctness of queries, but much more importantly we care about correctness of results. And those can't be tested without some DB in the loop.

> I've written plenty of unit tests in Rails and Flask (python), and you never mock the database. You just use a test variant of it, and the framework resets the database on every test. Yes, it takes a bit longer, but the tests are much simpler (no mocks) and you have more confidence that the DB code will run correctly in production.

Same experience in my background (Rails and Java/Hibernate).

> So, no, we don't have any tests that cover this.

How about extending pkg/cache/sql/integration_test.go

Those already run a fake Kubernetes environment and have the DB in the loop.

pkg/cache/sql/informer/listoption_indexer.go
richard-cox: Thinking about yaml an empty label value `foo: ` is provided as `foo: ""` via api (json). Not sure that makes any difference?

I couldn't find anything specific in the label selector for 'without this label' / 'with a label but NULL', though the generic utility of filter by everything that doesn't have a property value might be useful at some point

pkg/cache/sql/informer/listoption_indexer.go
richard-cox: There needs to be a way to differentiate between partial and exact matches. As long as that's there am open to offers on syntax (UI changes would be straight forward)

pkg/cache/sql/informer/listoption_indexer.go
richard-cox: As per slack message, this approach i think should work but would like to prod it with some real life queries.. so awaiting a dev image

pkg/cache/sql/informer/listoption_indexer.go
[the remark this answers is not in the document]
ericpromislow: I guess I could test the `after-upsert` handler directly. But I'm going to leave this, because better testing would involve an actual database, and then triggering addition of an object and verifying the expected labels have been added after. So no more work to be done here.

pkg/cache/sql/informer/listoption_indexer.go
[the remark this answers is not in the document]
ericpromislow: I changed the cast in `addLabels` so it now fails.

And what you're missing is: I forgot to push my changes on Monday (yesterday) :(

pkg/cache/sql/informer/listoption_indexer.go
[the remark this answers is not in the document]
ericpromislow: I looked at this, but it's simpler to create `countQuery` based on the first query, which refers only to tables and fields and doesn't have any of the pagination stuff.

Let's try building the final `QueryInfo` block a bit at a time, which should give us both more simplicity and clarity.

In fact I see I'm setting `queryInfo := &QueryInfo{}` at the start of `constructQuery` and then reassigning it. 
 `golangci-lint` doesn't flag that error, and does flag a lot of things that the `steve .golangci.json` says to ignore, so I don't think we need to add it to lasso right now.

Changing this required a few changes to the unit tests, mainly distinguishing undefined slices as nil vs an empty slice.

pkg/cache/sql/informer/listoption_indexer.go
moio: Number 1, let's focus back on fields (not labels) in this thread - discussion on labels, if needed, will be in a separate thread.

Number 2, @ericpromislow:

> my guess is that NULL/!NULL don't make sense for the non-labels, because either we index them (which means they can't be NULL, because they always come with a value, right?) or we don't index them, which means you can't query on those fields.

Hmm, my understanding is different. Assuming that we index a certain resource field, its value can very well be null (ie. missing or as explicit `null` in the YAML) - lots of structs corresponding to Kubernetes resources come with the "omitempty" directive for that reason.

We have to keep in mind that, right now, in such cases we simply [throw away the null and exchange it with an empty string](https://github.com/ericpromislow/lasso/blob/2f0f76ae5d300c21b1ae959a084256bb953984a0/pkg/cache/sql/informer/listoption_indexer.go#L197). That is arguably not really awesome, but it suffices when all we care for is string search, which is probably why this shortcut was originally taken. Now...


Number 3, @richard-cox: do I understand correctly that, from a UI perspective, we do not really need anything other than exact/inexact text matching for non-label properties, at least for now?

pkg/cache/sql/informer/listoption_indexer.go
moio: @richard-cox please note that the context for this thread is labels.

My understanding is that, for labels, you need power to match [label selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/).

Now label selectors do implement equality, which we call exact for normal fields. They do not implement "substring equality".

Can you confirm we will continue to need both "exact" and "substring" equality for normal fields, and label selector power on label fields, which _does not include "substring" equality_?

richard-cox:
```
Can you confirm we will continue to need both "exact" and "substring" equality for normal fields, and label selector power on label fields, which does not include "substring" equality?
```

@moio correct.

Label selectors are exact/not exact or in/not in a set. Filtering outside of labels would need to cover exact/not exact as well as substrings ... but not sets

pkg/cache/sql/informer/listoption_indexer.go
moio: > The SQL for the second call was this: 
> 
> ```
> SELECT o.object, o.objectnonce, o.dekid FROM "_v1_Event" o
>   JOIN "_v1_Event_fields" f ON o.key = f.key
>   JOIN "_v1_Event_labels" lt ON o.key = lt.key
>   WHERE
>     (f."metadata.name" = "lima-rancher-desktop.17e053aa93e89c71") AND
>     (lt.label = "fish" AND lt.value LIKE "%car% ESCAPE '\')
>   ORDER BY f."metadata.namespace" ASC, f."metadata.name" ASC 
>   LIMIT 100000
> ```
> 
> and it behaved as expected.


I tried the same but the behavior I got does not match yours for some reason. (One thing that does not match is the use of double quotes around `"fish"` and `"%car% ESCAPE '\'` - not sure if that is the culprit)


Please try with my DB, obtained by running Rancher briefly with the latest version of this PR: https://drive.google.com/drive/folders/1xIRN5fZ4ye7OD6eHsO-tTOYkOD6RaT5J?usp=sharing


Like in your case, events have no labels:

Query:
```sql
SELECT COUNT(DISTINCT e.key) AS events_with_labels
 FROM _v1_Event e
 WHERE (SELECT COUNT(*) FROM _v1_Event_labels l WHERE e.key = l.key) > 0;
```

Result:
```
0
```

Query:
```sql
SELECT COUNT(DISTINCT e.key) AS events_without_labels
FROM _v1_Event e
WHERE (SELECT COUNT(*) FROM _v1_Event_labels l WHERE e.key = l.key) = 0;
```

Result:
```
228
```

Running the query above (with fixed quotes) returns no results:
```sql
SELECT o.object, o.objectnonce, o.dekid FROM "_v1_Event" o
                                                 JOIN "_v1_Event_fields" f ON o.key = f.key
                                                 JOIN "_v1_Event_labels" lt ON o.key = lt.key
WHERE
    (f."metadata.name" = "lima-rancher-desktop.17e053aa93e89c71") AND
    (lt.label = "fish" AND lt.value LIKE '%car%' ESCAPE '\')
ORDER BY f."metadata.namespace" ASC, f."metadata.name" ASC
  LIMIT 100000;
```

Result:
```
```

And that is correct, as no fish nor car labels exists.

Problem is removing the `WHERE` conditions which is what would happen if there was no filter applied:

```sql
SELECT o.object, o.objectnonce, o.dekid FROM "_v1_Event" o
                                                 JOIN "_v1_Event_fields" f ON o.key = f.key
                                                 JOIN "_v1_Event_labels" lt ON o.key = lt.key
ORDER BY f."metadata.namespace" ASC, f."metadata.name" ASC
  LIMIT 100000;
```

Result:
```
```

That is not correct, as all 228 events should have returned. `JOIN`ing on fields filtered them away, in fact removing the JOIN brings the rows back:

```sql
SELECT o.object, o.objectnonce, o.dekid FROM "_v1_Event" o
                                                 JOIN "_v1_Event_fields" f ON o.key = f.key
ORDER BY f."metadata.namespace" ASC, f."metadata.name" ASC
  LIMIT 3;
```

Result:
```
%
 Unstructured||0
%
 Unstructured||0
%
 Unstructured||0
```


So I continue to think that `JOIN`ing on labels is not what we should do. Filtering on subqueries (something like `WHERE... AND (SELECT COUNT(*) FROM labels l WHERE l.key = o.key) > 0` seems to me to be the best way forward.

pkg/cache/sql/informer/listoption_indexer.go
moio: @ericpromislow as clarified in https://github.com/rancher/lasso/pull/117#discussion_r1869610527, we do not really need NULL/NOT NULL filtering for non-label fields for now.

How about just not supporting that operator and returning an error, until we decide it's actually needed and then we add support for it?

pkg/cache/sql/informer/listoption_indexer.go
moio: @ericpromislow please keep in mind the UI needs to evaluate results of a selector from a Kubernetes resource "as it is" (eg. taking a .spec.selector out of a Deployment, pass it to Steve to get a list of targeted pods in order to display a preview - If the user interactively changes the selector, update the preview on-the-fly).

Therefore there should be a way to pass in a plain Kubernetes selector as a filter with the exact same syntax and get the exact same semantics in return.

It is my understanding from Steve code that we are now parsing `filter=` query parameters with a syntax similar to vanilla Kubernetes selectors, with an extension for quoted strings (which are interpreted as exact matches, while the default is substring matching). Note that is different from standard Kubernetes selectors, where matching is exact by default (and substring matching does not even exist).

If the above is right (please correct me otherwise!) you might want to consider adding a hard distinction between `filter=` parameters - to be kept working exactly as today, and a new kind of parameters - say, `selectorFilter=` - that recognizes and implements Kubernetes selector syntax and semantics 1:1.

Please let me know what you think

pkg/cache/sql/informer/listoption_indexer.go
ericpromislow: Sure, this is now pushed for both lasso and steve (steve parses, lasso complains if a filter object somehow reaches it calling for a non-label existence check, but this shouldn't happen in the field).

pkg/cache/sql/informer/listoption_indexer.go
ericpromislow: It is correct that if the target of a `=` or `!=` operator is quoted, we do an exact match. Otherwise we do a substring match. This is an extension of the standard kubectl label selector parser, which is why I forked and modified it.

I'd rather not make any more changes until Richard is able to integrate these changes and see if they work as expected from the UI. I don't know if the UI in 2.10 releases is able to issue these kinds of filters -- I haven't seen any evidence of it.

pkg/cache/sql/informer/listoption_indexer.go
ericpromislow: First, correcting the query (for anyone who wants to follow along at home):
```
SELECT o.object, o.objectnonce, o.dekid FROM "_v1_Event" o
  JOIN "_v1_Event_fields" f ON o.key = f.key
  JOIN "_v1_Event_labels" lt ON o.key = lt.key
  WHERE
    (f."metadata.name" = "lima-rancher-desktop.17e053aa93e89c71") AND
    (lt.label = "fish" AND lt.value LIKE "%car%" ESCAPE '\')
  ORDER BY f."metadata.namespace" ASC, f."metadata.name" ASC 
  LIMIT 100000
```

And this does return no rows, as expected.

ericpromislow: Next, regarding this query:
```
SELECT o.object, o.objectnonce, o.dekid FROM "_v1_Event" o
                                                 JOIN "_v1_Event_fields" f ON o.key = f.key
                                                 JOIN "_v1_Event_labels" lt ON o.key = lt.key
ORDER BY f."metadata.namespace" ASC, f."metadata.name" ASC
  LIMIT 100000;
  ```

I only do a join on `X_labels` if there's an actual label query going on. This isn't the case,
so this query should never be executed. If you can provide a selector string that results in
this query, please let me know and I'll add it to the unit tests for `constructQuery`.

Comment (moio) [reply pkg/cache/sql/informer/listoption_indexer.go:288]: What about a != label filter though?

https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#equality-based-requirement says:

```
environment = production
tier != frontend
```

> The former selects all resources with key equal to environment and value equal to production. The latter selects all resources with `key` equal to `tier` and value distinct from `frontend`, **and all resources with no labels with the `tier` key.**

(emphasis mine)

File: pkg/cache/sql/informer/listoption_indexer.go (hunk 11 above)
Comment (richard-cox) [reply pkg/cache/sql/informer/listoption_indexer.go:631]: From a UI perspective splitting these into two non-overlapping filters would work. One covering the existing `filter` with our own syntax and another covering the kube label select (and standard kube filtering?) api syntax. It feels like the latter is a more neater solution, though i haven't looked through the changes in this PR.

The UI effort to cover either approach (single uber filter or two filters) is small.

Eric you're right, there's currently no way to exercise the label selector via api in the ui. Given an image with your changes in (that supports amd/x86) i could implement the feature in a single area and knock up a build that can very easily be used (by changing two rancher management.cattle.io `setting` values)

File: pkg/cache/sql/informer/listoption_indexer.go (hunk 12 above)
Comment (ericpromislow) [reply pkg/cache/sql/informer/listoption_indexer.go:288]: Actually, I implemented `NOT EXISTS` for labels incorrectly, and didn't catch that second case for the `labels != value` form, so they're both wrong.

**Comment by ericpromislow** [comment]
Two comments on the `!=` issue:

1. My mistake completely on the submitted `LABEL NOT-EXISTS` solution. It simply doesn't work. I've done live tests with a sqlite DB with the new form, and the results look correct. It also conceptually makes more sense.

2. Yes, when you're doing a NOT-EXISTS or != test with labels, you do need to do a `LEFT OUTER JOIN` with the main field rows on the left and the labels table on the right. The `LEFT OUTER JOIN` is only needed when doing these two comparisons -- in all other cases, if a particular key has no associated labels, there's no need to include it in the table we're doing processing on.

3. After all that, I didn't see a need for a `DISTINCT` qualifier. I'm now doing a `SELECT key FROM x JOIN x_labels WHEN key NOT IN (SELECT o2.key FROM x LEFT OUTER o2 JOIN x_labels l2 WHEN l2.label = NAME (AND l2.value = VALUE))`
and the inner query never returns duplicates.

As for performance on large sets, I assume that sqlite is smart enough to convert the inner query into a stream of rows that the outer query processes in a separate coroutine or thread. I've never read the source, but I've done tests (with Python code) and it was much faster to give sqlite a 12-line query to process than to give it a conceptually simpler query and have the Python code do further filtering on it.

**Comment by ericpromislow** [comment]
Biking home (in the rain) I realized the current solution still falls short for negative tests on labels in (at least) three ways:

1. We need to do the non-existence test on label `foo` for `metadata.labels.foo NOTIN (...)`

2. `hasNegativeExistenceTest` needs to also look for `NOTIN` operator on labels

3. `hasNegativeExistenceTest` should also verify that the `!=`, `NOT-EXISTS` and `NOT-IN` operators are being used with a `metadata.labels.X` subject. Otherwise there's no need to do an OUTER JOIN here.

File: pkg/cache/sql/informer/listoption_indexer.go [hunk 19]
```diff
@@ -191,25 +242,66 @@ func (l *ListOptionIndexer) afterDelete(key string, tx db.TXClient) error {
...
 	query += "\n  "
-	query += fmt.Sprintf(`JOIN "%s_fields" f ON o.key = f.key`, db.Sanitize(l.GetName()))
+	query += fmt.Sprintf(`JOIN "%s_fields" f ON o.key = f.key`, dbName)
+	if hasLabelFilter(lo.Filters) {
+		query += "\n  "
+		leftOuterPart := ""
+		if hasNegativeExistenceTest(lo.Filters) {
+			leftOuterPart = "LEFT OUTER "
+		}
+		query += fmt.Sprintf(`%sJOIN "%s_labels" lt ON o.key = lt.key`, leftOuterPart, dbName)
```
moio, line 295:
```suggestion
		query += "\n  "
		query += fmt.Sprintf(`LEFT OUTER JOIN "%s_labels" lt ON o.key = lt.key` dbName)
```

Rationale: `LEFT OUTER JOIN`'s results are the same as the default (`INNER`) `JOIN` - only if object rows have no labels `LEFT OUTER JOIN` returns more - filled in with `NULL`s.

Those extra rows will anyway be filtered away by the label conditions in all cases but non-existence, which is what we want.

So if my calculations are right we can simplify this code and let the SQL optimizer do the rest by always using the same `JOIN` type.

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -191,25 +242,66 @@ func (l *ListOptionIndexer) afterDelete(key string, tx db.TXClient) error {
...
+func (l *ListOptionIndexer) constructQuery(lo ListOptions, partitions []partition.Partition, namespace string, dbName string) (*QueryInfo, error) {
+	queryInfo := &QueryInfo{}
+
+	// First, what kind of filtering will we be doing?
 	// 1- Intro: SELECT and JOIN clauses
-	query := fmt.Sprintf(`SELECT o.object, o.objectnonce, o.dekid FROM "%s" o`, db.Sanitize(l.GetName()))
+	// There's a 1:1 correspondence between a base table and its _Fields table
+	// but it's possible that a key has no associated labels, so if we're doing a
+	// non-existence test on labels we need to do a LEFT OUTER JOIN
+	query := fmt.Sprintf(`SELECT o.object, o.objectnonce, o.dekid FROM "%s" o`, dbName)
```
moio, line 286: I see rows being duplicated when an object has multiple rows.

From [the data set shared before](https://drive.google.com/drive/folders/1xIRN5fZ4ye7OD6eHsO-tTOYkOD6RaT5J?usp=sharing):


Check that namespaces starting with `cattle` have multiple abels:
```sql
SELECT o.key, lt.label FROM "_v1_Namespace" o
    LEFT OUTER JOIN "_v1_Namespace_labels" lt ON o.key = lt.key
    WHERE o.key LIKE 'cattle%'
    ORDER BY o.key, lt.label;
```

Result: generally they do
| o.key                           | lt.key                               |
|---------------------------------|--------------------------------------|
| cattle-fleet-system             | field.cattle.io/projectId            |
| cattle-fleet-system             | kubernetes.io/metadata.name          |
| cattle-fleet-system             | objectset.rio.cattle.io/hash         |
| cattle-global-data              | cattle.io/creator                    |
| cattle-global-data              | field.cattle.io/projectId            |
| cattle-global-data              | kubernetes.io/metadata.name          |
| cattle-global-nt                | cattle.io/creator                    |
| cattle-global-nt                | field.cattle.io/projectId            |
| cattle-global-nt                | kubernetes.io/metadata.name          |
| cattle-impersonation-system     | authz.cluster.cattle.io/impersonator |
| cattle-impersonation-system     | cattle.io/creator                    |
| cattle-impersonation-system     | field.cattle.io/projectId            |
| cattle-impersonation-system     | kubernetes.io/metadata.name          |
| cattle-provisioning-capi-system | kubernetes.io/metadata.name          |
| cattle-system                   | field.cattle.io/projectId            |
| cattle-system                   | kubernetes.io/metadata.name          |
| cattle-ui-plugin-system         | kubernetes.io/metadata.name          |


Now, change the `SELECT` clause to actually return columns we are interested in this PR (and adding `o.key` because extremely long hexadecimal strings are hard for humans to read):

```sql
SELECT o.key, o.object, o.objectnonce, o.dekid FROM "_v1_Namespace" o
    LEFT OUTER JOIN "_v1_Namespace_labels" lt ON o.key = lt.key
    WHERE o.key LIKE 'cattle%'
    ORDER BY o.key, lt.label;
;
```

Results:
| o.key                           | o.object  | o.objectnonce | o.dekid |
|---------------------------------|-----------|---------------|---------|
| cattle-fleet-clusters-system    | 0x257F... |               | 0       |
| cattle-fleet-clusters-system    | 0x257F... |               | 0       |
| cattle-fleet-clusters-system    | 0x257F... |               | 0       |
| cattle-fleet-system             | 0x257F... |               | 0       |
| cattle-fleet-system             | 0x257F... |               | 0       |
| cattle-fleet-system             | 0x257F... |               | 0       |
| cattle-global-data              | 0x257F... |               | 0       |
| cattle-global-data              | 0x257F... |               | 0       |
| cattle-global-data              | 0x257F... |               | 0       |
| cattle-global-nt                | 0x257F... |               | 0       |
| cattle-global-nt                | 0x257F... |               | 0       |
| cattle-global-nt                | 0x257F... |               | 0       |
| cattle-impersonation-system     | 0x257F... |               | 0       |
| cattle-impersonation-system     | 0x257F... |               | 0       |
| cattle-impersonation-system     | 0x257F... |               | 0       |
| cattle-impersonation-system     | 0x257F... |               | 0       |
| cattle-provisioning-capi-system | 0x257F... |               | 0       |
| cattle-system                   | 0x257F... |               | 0       |
| cattle-system                   | 0x257F... |               | 0       |
| cattle-ui-plugin-system         | 0x257F... |               | 0       |


Can you reproduce those results? Could it be I am still missing something?

moio:
>     1. My mistake completely on the submitted `LABEL NOT-EXISTS` solution. It simply doesn't work. I've done live tests with a sqlite DB with the new form, and the results look correct. It also conceptually makes more sense.

Nah, no worries. That's what reviews are for - that is, when they do work! 😇 glad it's fixed now

>     2. Yes, when you're doing a NOT-EXISTS or != test with labels, you do need to do a `LEFT OUTER JOIN` with the main field rows on the left and the labels table on the right. The `LEFT OUTER JOIN` is only needed when doing these two comparisons -- in all other cases, if a particular key has no associated labels, there's no need to include it in the table we're doing processing on.

Ack!

>     3. After all that, I didn't see a need for a `DISTINCT` qualifier. I'm now doing a `SELECT key FROM x JOIN x_labels WHEN key NOT IN (SELECT o2.key FROM x LEFT OUTER o2 JOIN x_labels l2 WHEN l2.label = NAME (AND l2.value = VALUE))`
>        and the inner query never returns duplicates.

Hmm, I see duplicates though. Taking this discussion to the review to keep it in one place.

> As for performance on large sets, I assume that sqlite is smart enough to convert the inner query into a stream of rows that the outer query processes in a separate coroutine or thread. I've never read the source, but I've done tests (with Python code) and it was much faster to give sqlite a 12-line query to process than to give it a conceptually simpler query and have the Python code do further filtering on it.

Absolutely, we should default to defer complexity to the SQLite query planner as much as possible. We should complicate queries only if absolutely needed, and I have no reasons to believe this is a concern for now.

Thanks!

moio (changes requested):

pkg/cache/sql/informer/listoption_indexer.go
ericpromislow: Yes, I can definitely reproduce it, and here's the sql I used that gives clearer output (splitting the serialized object blob and splitting to show the start and end, because they all have the same start):
```
sqlite> SELECT o.key,
   ...>        substr(hex(o.object), 1, 10),
   ...>        substr(hex(o.object), length(hex(o.object)) - 10),
   ...>        substr(hex(o.objectnonce), 1, 10),
   ...> substr(hex(o.dekid), 1, 10) FROM "_v1_Namespace" o LEFT OUTER JOIN "_v1_Namespace_labels" lt ON o.key = lt.key WHERE lt.label like "%/%" ORDER BY o.key, lt.label;
cattle-fleet-clusters-system|257F030101|97374656D00||30
cattle-fleet-clusters-system|257F030101|97374656D00||30
cattle-fleet-clusters-system|257F030101|97374656D00||30
cattle-fleet-local-system|257F030101|E6574657300||30
cattle-fleet-local-system|257F030101|E6574657300||30
cattle-fleet-local-system|257F030101|E6574657300||30
…
```

So, yes, it needs a "DISTINCT" to remove the duplicates.

moio: OK, I am resolving this conversation now.

I'll be waiting on @richard-cox and colleagues to test various label filtering and keep an eye on whether any duplicates are ever returned (they should not).

Also as a note in the future: should this approach not work well (because DISTINCT adds some overhead) one possibility is to not JOIN labels at all and only look for them in subqueries in the WHERE clause (like it's done for the NOT IN case). This might or might not be beneficial for performance - but let's leave that optimization for later, if needed.

Thanks!

moio:
Tentative :thumbsup: from my side - let's hear @richard-cox's comments testing this.

I'd also suggest a second look from a Frameworks colleague - @tomleb maybe?

richard-cox:
Had a chance to look at the image / feature today.

Comment WIP, updating.

1. With no params provided it looks like some filtering is happening anyway 
  a) image
  b) image
  c) image
2. We may need a feature to flip reverse the resources
  a) Need to determine the services associated with a deployment
  b) currently this happens by fetching the pods associated with a deployment and then iterating over all services and testing the service's pod selector against the deployments pods

ericpromislow:
@Tom re https://github.com/rancher/lasso/pull/117#pullrequestreview-2532366126

Before I do that I created three slightly different configmaps (I like to use readable words that are different enough):
```
configmap01.yaml:  labels:
configmap01.yaml-    cm01a: snail
configmap01.yaml-    cm01b: shade
--
configmap02.yaml:  labels:
configmap02.yaml-    cm02a: theta
configmap02.yaml-    cm02b: raspy
--
configmap03.yaml:  labels:
configmap03.yaml-    cm03e: froze
configmap03.yaml-    cm03f: lever
```

Referring to kubernetes:
```
$ k get configmap -l cm01a=snail,cm02b=raspy
No resources found in default namespace.  # I thought this is an 'OR', but it looks like an 'AND'

$ k get configmap -l cm01a=snail -l cm02b=raspy # Here the last one wins
NAME   DATA   AGE
cm02   1      15m 

$ k get configmap -l cm02b=raspy -l cm01a=snail # Last one wins again
NAME   DATA   AGE
cm01   2      17m
```

And with no changes to the parser and sql generator:
```
$ curl -skL 'https://localhost:5111/v1/configmaps?filter=metadata.labels.cm01a=snail,metadata.labels.cm02b=raspy' | jq .count
2
```

The generated sql was:
```
SELECT FROM ... "_v1_ConfigMap" o
  JOIN "_v1_ConfigMap_fields" f ON o.key = f.key
  LEFT OUTER JOIN "_v1_ConfigMap_labels" lt ON o.key = lt.key
  WHERE
    ((lt.label = "cm01a" AND lt.value LIKE "%snail1% ESCAPE '\') OR (lt.label = "cm02b" AND lt.value LIKE "%raspy%"))
```
Now I'm confused...  Did I get the semantics of `filter=F1&filter=F2....&filter=` and `filter=F1,F2,...` reversed?

I think so.  Here's how the upstream code parses a group of comma-separated expressions:
```
// Matches for a internalSelector returns true if all
// its Requirements match the input Labels. If any
// Requirement does not match, false is returned.
func (s internalSelector) Matches(l Labels) bool {
	for ix := range s {
		if matches := s[ix].Matches(l); !matches {
			return false
		}
	}
	return true
}
```

And I'm mapping them to an OR-group. End of day here, but this will need correcting tomorrow.

ericpromislow:
Tom was right (as always :). I'm now doing self-joins on label tables, even if there's only one instance of a label in the compound query (which simplifies the code generator), and live probing makes sense.

tomleb (approved):
LGTM. I've done some light testing on things that were failing before and they now work correctly.

PR #118, opened 2024-11-12
Ensure all failed transactions are rolled back.

ericpromislow:
Related to [#46410](https://github.com/rancher/rancher/issues/46410)

I only found one instance.

Basically any function that receives a transaction handle and returns a
non-nil error is responsible for ensuing the transaction is rolled back.
This is covered by any failed calls to `tx.Exec` or `tx.StmtExec` but
in this case `ListOptionIndexer#afterUpsert` can fail on other situations,
and is therefore responsible for rolling back the transaction.


moio (approved):

gehrkefc (approved):
LGTM

PR #122, opened 2025-01-13
Replace primary/secondary sort fields with an array of sort directives.

ericpromislow:
This allows collection URLs to have a sort query that sorts on more than two fields

The use of `PrimaryField` and `SecondaryField` is too limiting, and leads to unlikely edge-cases,
like having a Secondary Field but no Primary Field. Better to just maintain parallel arrays of 
fields and sort-orders.

Related to [#48092](https://github.com/rancher/rancher/issues/48092)

### DIFF
--- a/pkg/cache/sql/informer/listoption_indexer.go
+++ b/pkg/cache/sql/informer/listoption_indexer.go
@@ -382,6 +382,9 @@ func (l *ListOptionIndexer) constructQuery(lo ListOptions, partitions []partitio
 	}
 
 	// 3- Sorting: ORDER BY clauses (from lo.Sort)
+	if len(lo.Sort.Fields) != len(lo.Sort.Orders) {
+		return nil, 0, "", fmt.Errorf("sort fields length %d != sort orders length %d", len(lo.Sort.Fields), len(lo.Sort.Orders))
+	}
 	if len(lo.Sort.Fields) > 0 {
 		orderByClauses := []string{}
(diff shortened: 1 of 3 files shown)


moio (approved):

pkg/cache/sql/informer/listoption_indexer.go
```diff
@@ -284,33 +284,20 @@ func (l *ListOptionIndexer) ListByOptions(ctx context.Context, lo ListOptions, p
...
+			columnName := toColumnName(field)
+			if err := l.validateColumn(columnName); err != nil {
+				return nil, 0, "", err
+			}
 
-		direction := "ASC"
-		if lo.Sort.SecondaryOrder == DESC {
-			direction = "DESC"
+			direction := "ASC"
+			if lo.Sort.Orders[i] == DESC {
```
tomleb, line 296: We should check that len(lo.Sort.Fields) == len(lo.Sort.Orders) and return an error, otherwise we (although unlikely) risk panicking.

ericpromislow:
This now depends on lasso PR [117](https://github.com/rancher/lasso/pull/117) in order to improve the unit tests for `ListByOptions`

pkg/cache/sql/informer/listoption_indexer.go
ericpromislow: Done

moio (approved):

tomleb (approved):

PR #123, opened 2025-01-16
Remove sqlcache

tomleb:
As lightly discussed, the vai cache is currently very steve specific and it's going to be faster to iterate if it's embedded directly into the steve repo.

This removes all mentions of the Vai cache.

Draft until we actually have the sqlcache added to steve.


moio (approved):
LGTM

ericpromislow (approved):
I've been using this and it's fine

PR #156, opened 2025-04-02
Add "ctx" label to lasso handler metrics

aruiz14:
Similarly to how it was added to `lasso_controller_total_cached_object` in #95, having this label will allow us to differentiate handler executions from different contexts while analyzing Prometheus metrics. It reuses the same instrumentation added for identifying the caches: https://github.com/rancher/rancher/pull/46784.

This change should be backwards compatible with existing dashboards, as they must be already using `sum` in some way.

Some example queries using a local build:
- Top 10 downstream handler executions:
Screenshot 2025-04-02 at 11 12 54
- Comparing handlers from 2 downstream with 3 nodes vs. 1 node respectively:
Screenshot 2025-04-02 at 10 27 43

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -50,6 +50,7 @@ type controller struct {
 	startLock sync.Mutex
 
 	name        string
+	ctxID       string
 	workqueue   workqueue.RateLimitingInterface
 	rateLimiter workqueue.RateLimiter
 	informer    cache.SharedIndexInformer
@@ -174,6 +175,7 @@ func (c *controller) Start(ctx context.Context, workers int) error {
 		return fmt.Errorf("failed to wait for caches to sync")
 	}
 
+	c.ctxID = metrics.ContextID(ctx)
 	go c.run(workers, ctx.Done())
 	c.started = true
 	return nil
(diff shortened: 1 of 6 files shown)


pkg/controller/controller.go
```diff
@@ -50,6 +50,7 @@ type controller struct {
 	startLock sync.Mutex
 
 	name        string
+	ctxID       string
```
tomleb, line 53: Should this be behind a lock? It's written once when creating the controller object, it's written a second time in `Start()` and we read it in another goroutine.

tomleb (commented):
Nice addition, should certainly be useful

moio:
@aruiz14 as you know I am onboard with the general plan, but I am no longer the right person to give code reviews generally

pkg/controller/controller.go
aruiz14: It's already behind the `startLock` isn't it? And will only be assigned once (as `Start()` returns early if `c.started` is `true`. Also, the code that reads its value is not yet running, so we should be fine there too.

ericpromislow (approved):
LGTM

pkg/controller/controller.go
tomleb: You're right. In that case the write is done under a lock and the goroutine doing the read can never read the value while it's being written to.

PR #159, opened 2025-05-06
[main] Allow for partial discovery list

joshmeranda:
https://github.com/rancher/rancher/issues/49299

Allow for partial GVK discovery and report errors via logs.

Note: The `APIService` startup is a bit bumpier. Since the discovery is allowed to succeed before the imperative-api is up an running, it can take more time before the `APIService` is able to discover the new resources and allow users to access those resource via ther kube apiserver.


ericpromislow (approved):
Makes sense

PR #160, opened 2025-05-07
Avoid race-condition removing object with multiple finalizers

aruiz14:
Refers to rancher/rancher#49328


pkg/controller/controller_test.go
```diff
@@ -0,0 +1,96 @@
...
+	"k8s.io/utils/ptr"
+)
+
+//go:generate mockgen --build_flags=--mod=mod -package controller -destination ./mock_shared_index_informer_test.go k8s.io/client-go/tools/cache SharedIndexInformer
+
+func TestController_multiple_finalizers_race(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	queue := workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
```
ericpromislow, line 24: These calls are now deprecated in k8s/client-go v0.32 but dang if I can figure out how to use the generic equivalents.

aruiz14: Correct, it's actually [`type controller struct`](https://github.com/rancher/lasso/blob/main/pkg/controller/ controller.go#L53) which has some field with a deprecated type. Happy to help refactoring that to adopt new types based on generics in a separate PR.

aruiz14: After taking a quick look, and have some comments about this:
1. The new typed interface allows using `string` instead of `any`, getting stricter type-checking and avoiding checks like [this](https://github.com/rancher/lasso/blob/main/pkg/controller/controller.go#L212). 
   - (The deprecated `RateLimitingQueue` is just an "alias" of `TypedRateLimitingQueue[any]`
2. This leads to a breaking change, since [`RateLimiter` from `Options`](https://github.com/rancher/lasso/blob/main/pkg/controller/controller.go#L69) should also need to changed to `TypedRateLimiter[string]`.
3. Sadly, only the `workqueue` was adapted to support generics. It would be much more interesting if the [`SharedIndexInformer`](https://pkg.go.dev/k8s.io/client-go@v0.33.0/tools/cache#SharedInformer) and its [`Store`](https://pkg.go.dev/k8s.io/client-go@v0.33.0/tools/cache#Store) were also parameterized, so we don't need to cast objects to `runtime.Object`.

Taking that into account, I believe there is not much motivation to move this forward at the moment. We could still migrate to use `TypedRateLimitingQueue[any]`, which will keep the compatibility with existing code but also completely miss the benefit of using generics. Also, that could also even reduce our chances of ever migrating to use the correct type.

aruiz14: Here is a draft PR with the changes, for completeness:
https://github.com/rancher/lasso/pull/162
please feel free to do whatever you think appropriate with it

ericpromislow (approved):
It's fine. Not worth dealing with the deprecated functions for now

PR #162, opened 2025-05-08
breaking(controller): migrate workqueue to use strict typing with generics

aruiz14:


pkg/controller/controller.go
```diff
@@ -46,12 +46,14 @@ type Controller interface {
 	Start(ctx context.Context, workers int) error
 }
 
+type workqueueKey = string
```
aruiz14, line 49: This is not strictly needed, only added it for better visibility of the places where this type is used.
Exported fields/types already use `string` directly, otherwise we would need to make it public.

PR #164, opened 2025-06-23
add failing test

aruiz14:


crobby (approved):
Fix seems good.  Should definitely prevent the panic.  Bonus points for the quick test as well.

kinarashah (approved):

PR #175, opened 2025-08-28
Watcher's ResultChan must be properly closed after Stop is called

aruiz14:
According to `apimachinery`'s [`Watch` interface](https://github.com/kubernetes/apimachinery/blob/master/pkg/watch/watch.go#L29-L51) (emphasis mine):

>  	// ResultChan returns a channel which will receive events from the event
>  	// producer. If an error occurs or Stop() is called, **the producer must
>  	// close this channel and release any resources used by the watch**.
>  	// Closing the result channel tells the consumer that no more events will be
>  	// sent.
>  	ResultChan() <-chan Event

However, our implementation of this interface does not satisfy that (I added a failing test that confirms it), forcing users to drain `ResultChan` after calling `Stop` in order to prevent memory leaks.
This change should not have any side effects on existing code with the mentioned workaround (with the unlikely exception of those events being actually used during draining, which we could argue should be discouraged. In other words, users must not use `ResultChan` once they call `Stop`).

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -339,14 +339,21 @@ func (c *Client) injectKind(w watch.Interface, err error) (watch.Interface, erro
 	}()
 
 	return &watcher{
-		Interface: w,
+		wrapped:   w,
 		eventChan: eventChan,
 	}, nil
 }
 
 type watcher struct {
-	watch.Interface
+	wrapped   watch.Interface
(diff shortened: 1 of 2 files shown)


aruiz14:
Adding @brandond as suggested by @tomleb

tomleb (approved):

PR #185, opened 2026-02-18
Additional tests for lasso

crobby:
Issue: https://github.com/rancher/rancher/issues/45252

Adds integration tests for `SharedControllerFactory` that run against a real API server via [envtest](https://book.kubebuilder.io/reference/envtest), plus the tooling to install envtest reproducibly.

**Tests**
- `tests/integration/sharedcontrollerfactory_test.go` covers pod event delivery, syncing only changed objects, and `ForObject`. Each test starts its own control plane and runs in parallel.
- New `make test-integration` target, wired into CI. `make test` now scopes to `./pkg/...`.

**envtest tooling**
- `setup-envtest` is pinned in `gotools/setup-envtest/go.mod` and invoked via `go tool` — no `go install ...@latest`, so the version is reproducible and Renovate-managed like any other dependency.
- `scripts/install-envtest.sh` pins the control plane binaries (`v1.36.2`) with a SHA-512 checksum per platform, verifies the download, and sideloads it. Skips the download if the version is already in the local store.
- Guard: fails fast if the envtest minor version doesn't match `k8s.io/client-go`, so we don't test against an API server one minor off from the client.
- `scripts/bump-envtest.sh` rewrites the version and all checksums from the upstream release manifest when client-go moves, instead of copying hashes by hand.

**Docs**
- `tests/integration/README.md` — how to run, how the install works, why it's pinned, how to bump.
- `gotools/README.md` — the one-module-per-tool convention.
- Root README gains the `Developing` / `Running Tests` sections the table of contents already linked to but which didn't exist.


ericpromislow (changes requested):
From the issue, "In the ci, all tests running using go test should run with the -race flag."

Adding `-test` to `scripts/test*.sh` passed so maybe add that?

Oops. I meant "Adding `-race` …"

crobby:
> From the issue, "In the ci, all tests running using go test should run with the -race flag."
> 
> Adding `-test` to `scripts/test*.sh` passed so maybe add that?
> 
> Oops. I meant "Adding `-race` …"

That's a fine point.  Will do.

ericpromislow (approved):
Looks good. That test file is great

PR #191, opened 2026-03-31
[2.15] Use a sha-pinned version of setup-envtest

ericpromislow:
Ref [#1899](https://github.com/rancher/rancher-security/issues/1899)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,7 +19,8 @@ jobs:
       with:
         go-version-file: 'go.mod'
     - name: Install env-test
-      run: go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+      run: go install sigs.k8s.io/controller-runtime/tools/setup-envtest@8122a6266696b980fc9219c2fbb48bb4ce60d12b #v0.23.2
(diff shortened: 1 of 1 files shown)


macedogm (approved):

tomleb (changes requested):
To keep consistency with the other repository (steve), what do you think of using this script to install setup-envtest? https://github.com/rancher/steve/blob/main/scripts/install-envtest.sh

The script also checksums the binaries (apiserver / etcd) that are being installed by setup-envtest.

EDIT: Oops, meant this to be a comment

crobby:
@ericpromislow  Should we try to get this in?

ericpromislow:
Regarding @tomleb's suggestion, I can either run `install-envtest` in steve and copy/paste the workflow lines from steve into lasso, or I can copy the scripts from lasso into steve and run them there. But then I have to copy `steve/gotools/setup-envtest/` as well, and have to keep the SHAs in lasso in sync with steve, and the versions in lasso's `go.mod` in sync with steve's.  Sounds easier to just copy over the workflow command from steve.

crobby (approved):
Seems like a reasonable approach.

PR #197, opened 2026-04-29
Revert "Add option to disable WatchList feature"

rohitsuse:
Reverts rancher/lasso#186

Issue : https://github.com/rancher/rancher/issues/53703

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -21,11 +21,10 @@ const (
 )
 
 type Options struct {
-	Namespace        string
-	Resync           time.Duration
-	TweakList        TweakListOptionsFunc
-	WaitHealthy      func(ctx context.Context)
-	DisableWatchList bool
+	Namespace   string
+	Resync      time.Duration
+	TweakList   TweakListOptionsFunc
+	WaitHealthy func(ctx context.Context)
 }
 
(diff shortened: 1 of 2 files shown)


ericpromislow (approved):
The best PRs only delete code

gehrkefc (approved):

tomleb:
@rohitsakala Does this affect wrangler? I'm thinking backward compatibility-wise and its semver strategy. Just wanna ensure that we wouldn't be breaking one of the wrangler consumer (k3s / rke2 or watever)

rohitsuse:
> @rohitsakala Does this affect wrangler? I'm thinking backward compatibility-wise and its semver strategy. Just wanna ensure that we wouldn't be breaking one of the wrangler consumer (k3s / rke2 or watever)

I seached if the fields in this PR are being used in `wrangler`. Wrangler is not using the fields in this PR. So, there won't be breaking changes for wrangler.

tomleb:
@rohitsakala Right. I guess that ends up in a grey area where in order to use wrangler you need to call into lasso directly to create the factory, so someone might be relying on this, and when they'll update wrnagler, the lasso version will also upgrade and it's going to break. So I'm thinking it my be simpler to keep this in :thinking:

rohitsuse:
Closing this since it will break things.

